PR #15, opened 2023-04-16
Asset manager

kvark:
The goal here is to create a mechanism by which:
  - the assets can be loaded in 2 steps: first cooked to an optimized format, and then served to the user
  - this cooking result can be cached automatically between runs to fasten the startup
  - all the processing can be done in parallel using Choir

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 /Cargo.lock
 .DS_Store
+cooked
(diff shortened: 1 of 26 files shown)


kvark:
This is by no means done, but it's looking promising. Besides, at this point it accumulated so many changes I don't want to roll back.

PR #16, opened 2023-05-11
AssetManager v2

kvark:
Following up on the asset manager.
The main goal is to port all the scene loading to the new asset manager API.

sponza


kvark:
Need to rethink `SynCell` usage one more time...

PR #42, opened 2023-10-18
Support decoupled shading from reservoirs

kvark:
This is disabled by default because it clearly produces a lot of fireflies on the default scene:

decoupled-shading-fireflies


kvark:
getting this to work with pairwise MIS turns out to be challenging

PR #70, opened 2024-01-26
Fix architecture diagram link

berkus:
Make link absolute.


kvark (commented):
Thanks for the PR!
Why do you think the link needs to be absolute?
If I'm on a different branch, for example, wouldn't I want to see that branch's picture?

berkus:
It doesn't display the picture at all, otherwise.

PR #87, opened 2024-02-22
Allow tearing flag in the surface config

kvark:


kvark:
Have some concerns about this being the desired API. In particular, it may be required by the user to have vsync block for the acquire.

PR #92, opened 2024-03-04
Disable presentation on Intel when Nvidia is detected

### LINKED ISSUE #88: crashes on startup for PRIME enabled hardware thats configured to use the onboard nvidia not the intel
... seems to specifically choose the incorrect (intel) video card even tho 'prime-select query' output says 'nvidia'

temporary workaround, switch-around your prime configuration to use the *worse* 'intel" video card, reboot, and then it won't crash.
run: sudo prime-select intel
reboot

Edit: [vulkaninfo](https://github.com/kvark/blade/files/14471830/vi2.txt)

kvark:
Closes #88
Similar to https://github.com/gfx-rs/wgpu/pull/1898

### DIFF
--- a/blade-graphics/src/vulkan/init.rs
+++ b/blade-graphics/src/vulkan/init.rs
@@ -5,6 +5,21 @@ use ash::{
 use naga::back::spv;
 use std::{ffi, mem, sync::Mutex};
 
+mod db {
+    pub mod intel {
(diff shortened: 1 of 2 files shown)


davidbuzz:
I know its already merged... but I tested this, and it works for me.   
... here's the bunnymark example working now where it didn't before: 
image

Issue #93, opened 2024-03-11
Don't block use of Intel on Nvidia hybrid systems.


flukejones:
The recent commit https://github.com/kvark/blade/commit/30c4fa41f1476e932b9c1f104665b0a2904f94c4 makes it so that things like zed run on the dgpu full time, this is not an acceptable solution for https://github.com/kvark/blade/issues/88 as it causes excessive battery drain, heat, etc.

In that issue I reference https://github.com/gfx-rs/wgpu/pull/4110 because it appears to be a very similar use case. 

My own system is currently:

Operating System: Fedora Linux 40
Kernel Version: 6.8.0-rc7+ (64-bit)
Graphics Platform: Wayland
Processors: 32 × Intel® Core™ i9-14900HX
Memory: 62.4 GiB of RAM
Graphics Processor: Mesa Intel® Graphics
Manufacturer: ASUSTeK COMPUTER INC.
Product Name: ROG Strix SCAR 16 G634JYR_G634JYR_000045397
System Version: 1.0

Installed mesa version is: 24.0.0

Installed nvidia version: 550.54.14

I also have older laptops I can test. Plus I have tested on fedora 39 quite fine which used 6.6.x and 6.7.x kernels. The desktop is irrelevant here, I've tested on COSMIC, Gnome, KDE. What is of note however is that I do *not* use Xorg sessions and haven't done for years.

The proper and expected solution is to find the exact cause of https://github.com/kvark/blade/issues/88 and either fix that, or work around that one specific case. A blanket blocking of all intel/nvidia just handicaps everyone regardless.

# Logs

## Output from example using Intel:

The ` VK_ICD_FILENAMES=/usr/share/vulkan/icd.d/intel_icd.x86_64.json` wasn't actually required here.

```
❯ VK_ICD_FILENAMES=/usr/share/vulkan/icd.d/intel_icd.x86_64.json RUST_LOG=debug cargo run --example particle
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/examples/particle`
[2024-03-11T04:35:49Z DEBUG egui_winit::clipboard] Initializing arboard clipboard…
[2024-03-11T04:35:49Z DEBUG egui_winit::clipboard] Initializing smithay clipboard…
[2024-03-11T04:35:49Z WARN  blade_graphics::hal::init] Requested layer is not found: "VK_LAYER_KHRONOS_validation"
[...]

PR #94, opened 2024-03-13
Limit the Intel presentation bug to Xorg only

kvark:
Follow-up to #92 that narrows the workaround down a bit (to Xorg systems).
Helps #93 but doesn't completely address it.

### DIFF
--- a/blade-graphics/src/vulkan/init.rs
+++ b/blade-graphics/src/vulkan/init.rs
@@ -38,11 +38,16 @@ struct AdapterCapabilities {
     shader_info: bool,
 }
 
+struct SystemBugs {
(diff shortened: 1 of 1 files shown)


flukejones:
This is a good start for sure. Since it limits the impact to xorg, lets go with that.

Is blade able to determine which of the gpu is primary? If so that might be a way to narrow the case down to "xorg with nvidia dgpu as primary"

kvark:
I don't know how to determine the main adapter as seen by nvidia-prime.
Based on https://github.com/canonical/nvidia-prime/blob/587c5012be9dddcc17ab4d958f10a24fa3342b4d/prime-select#L56, we could check the contents of `/etc/prime-discrete`.

PR #95, opened 2024-03-13
vk: detect prime-select forced to Nvidia

kvark:
Follow-up to #94 that makes our check as strict as possible (within the data we have).
Closes #93 
See https://github.com/canonical/nvidia-prime/blob/master/prime-select for more info.

PR #97, opened 2024-03-18
vk: include Wayland in the Nvidia-Intel workaround

kvark:
Follow-up to #95, cc @flukejones

### DIFF
--- a/blade-graphics/src/vulkan/init.rs
+++ b/blade-graphics/src/vulkan/init.rs
@@ -36,9 +36,10 @@ struct AdapterCapabilities {
     shader_info: bool,
 }
 
+#[derive(Debug)]
 struct SystemBugs {
     /// https://gitlab.freedesktop.org/mesa/mesa/-/issues/4688
-    intel_unable_to_present_on_xorg: bool,
+    intel_unable_to_present: bool,
 }
 
 // See https://github.com/canonical/nvidia-prime/blob/587c5012be9dddcc17ab4d958f10a24fa3342b4d/prime-select#L56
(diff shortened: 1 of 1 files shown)


kvark:
@davidbuzz has confirmed that this PR picks Nvidia on their machine. Merging.

blade-graphics/src/vulkan/init.rs
```diff
@@ -341,14 +341,13 @@ impl super::Context {
...
-            Some((_, raw_window_handle::RawDisplayHandle::Xlib(_))) => true,
-            Some((_, raw_window_handle::RawDisplayHandle::Xcb(_))) => true,
-            _ => false,
-        };
         let bugs = SystemBugs {
-            intel_unable_to_present_on_xorg: is_xorg && is_nvidia_prime_forced(),
+            //Note: this is somewhat broad across X11/Wayland and different drivers.
+            // It could be narrower, but at the end of the day if the user forced Prime
+            // for GLX it should be safe to assume they want it for Vulkan as well.
+            intel_unable_to_present: is_nvidia_prime_forced(),
```
flukejones, line 348: This only applies to Xorg. Wayland completely ignores this, but using it as an indication of desired behaviour may work here.

flukejones: The general expected behaviour of desktop applications is that they default to igpu unless one of
```
PrefersNonDefaultGPU=false
X-KDE-RunOnDiscreteGpu=false
```
is set in the application.desktop file, or some desktops provide a method to offload with right-click menu.

However when it comes to games, typically vulkan defaults to dgpu. While opengl requires intervention.

PR #105, opened 2024-04-01
GLES explicit bindings

kvark:
Every backend has specifics on how the bindings are treated. But all of them eventually need to ensure the `naga::Module` bindings are filled up, and this process is backend-independent. Now it's done in a separate function.
With all the improvements to GLES, it's now able to run the "particle" example on desktop:
```bash
RUSTFLAGS="--cfg gles" CARGO_TARGET_DIR=./target-gl cargo run --example particle
```
particles-gles

TODO: check how it works with Zed

### DIFF
--- a/blade-egui/src/lib.rs
+++ b/blade-egui/src/lib.rs
@@ -302,16 +302,16 @@ impl GuiPainter {
             // Make sure clip rect can fit within an `u32`.
             let clip_min_x = (sd.scale_factor * clip_rect.min.x)
                 .clamp(0.0, sd.physical_size.0 as f32)
-                .trunc() as u32;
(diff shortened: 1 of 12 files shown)


kvark:
Was debugging Zed on GLES, and it's quite tough. The tools for GLES on Linux are quite poor.
RenderDoc shows stuff that *should* be working, but the driver thinks otherwise.
Specifically, I'm seeing that, for each vertex in the quad, the clip distances generated from VS are roughly from 0 to 600. However, inside the pixel shaders they come out negative in the Y component specifically.
I'm believe I'm just seeing a driver bug here (on the AMD 6850U).

Issue #106, opened 2024-04-06
"Memory block wasn't deallocated" is printed when exiting Zed builds on Linux


mikayla-maki:
A little sleuthing reveals that this error comes from the `gpu_alloc` crate, and is run when it's `MemoryBlock` type is dropped without calling `MemoryBlock::deallocate`. Probably because the `MemoryManager` in `blade-graphics/src/vulkan/mod.rs` isn't being cleared in this case.

kvark:
Following the comment of #107 - unless this is reproducible on the examples, it's likely a problem on the application side

PR #107, opened 2024-04-06
Deallocate GPU memory when dropping the Vulkan Context

mikayla-maki:
fixes https://github.com/kvark/blade/issues/106

### DIFF
--- a/blade-graphics/src/vulkan/resource.rs
+++ b/blade-graphics/src/vulkan/resource.rs
@@ -9,6 +9,20 @@ struct Allocation {
     handle: usize,
 }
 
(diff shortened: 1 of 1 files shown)


kvark (approved):
Thank you!

kvark (changes requested):
Awktually, I think this is a wrong fix?
Valid usage of Blade (as well as pure Vulkan) is to free memory before destroying the cotext.
So the problem is in blade_renderer.rs on Zed side, not here.

mikayla-maki:
Gotcha! Then I'll go look there, thank you :)

PR #115, opened 2024-05-04
Rank available devices by device type

VitorRamos:
Prefer discrete GPU over CPU device when possible

### DIFF
--- a/blade-graphics/src/vulkan/init.rs
+++ b/blade-graphics/src/vulkan/init.rs
@@ -368,14 +368,33 @@ impl super::Context {
             };
 
         let physical_devices = instance.core.enumerate_physical_devices().unwrap();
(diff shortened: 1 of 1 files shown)


VitorRamos:
Just learned we can use MESA_VK_DEVICE_SELECT to select the card and sorting will mess with that.

kvark:
I don't feel positive about messing with the device order unless there is no other way. Vulkan loader and ecosystem is built around the idea that an application should just grab the first available physical device if it doesn't particularly care. I.e. if you are a full-screen game, then you do care to pick a discrete GPU. If you are, say, a terminal window (ala Alacritty), you should use whatever GPU the user configures as default. `MESA_VK_DEVICE_SELECT` is one of the ways the user can control this.

VitorRamos:
Makes sense, closing this pr.

PR #125, opened 2024-06-02
gles: set unpack alignment to 1

lukaslihotzki:
Without this PR, GLES expects the default unpack alignment of 4 that seems not to be present on Vulkan. This causes, for example, a bug in Zed's glyph rendering (except for some glyphs which are 4-aligned randomly):
image


kvark (approved):
Thank you!

PR #126, opened 2024-06-02
gles: implement blending and write masks

lukaslihotzki:
Blending fixes the current mostly-transparent-window bug in zed-industries/zed#9581 (screenshot is attached there). Write masks weren't necessary, but easy to do.

### DIFF
--- a/blade-graphics/src/gles/command.rs
+++ b/blade-graphics/src/gles/command.rs
@@ -227,6 +227,17 @@ impl super::PassEncoder<'_, super::RenderPipeline> {
     ) -> super::PipelineEncoder<'b> {
         self.commands
             .push(super::Command::SetProgram(pipeline.inner.program));
+
+        match &pipeline.inner.color_targets[..] {
+            &[(blend_state, write_masks)] => self
+                .commands
+                .push(super::Command::SetAllColorTargets(blend_state, write_masks)),
+            separate => self.commands.extend(separate.iter().zip(0..).map(
+                |(&(blend_state, write_masks), i)| {
+                    super::Command::SetSingleColorTarget(i, blend_state, write_masks)
+                },
+            )),
+        }
         super::PipelineEncoder {
             commands: self.commands,
             plain_data: self.plain_data,
@@ -454,6 +465,38 @@ impl crate::VertexFormat {
             Self::I32Vec2 => (2, glow::INT),
             Self::I32Vec3 => (3, glow::INT),
             Self::I32Vec4 => (4, glow::INT),
+        }
+    }
+}
+
+impl crate::BlendFactor {
+    fn to_gles(self) -> u32 {
+        match self {
+            Self::Zero => glow::ZERO,
+            Self::One => glow::ONE,
+            Self::Src => glow::SRC_COLOR,
+            Self::OneMinusSrc => glow::ONE_MINUS_SRC_COLOR,
+            Self::SrcAlpha => glow::SRC_ALPHA,
+            Self::OneMinusSrcAlpha => glow::ONE_MINUS_SRC_ALPHA,
+            Self::Dst => glow::DST_COLOR,
+            Self::OneMinusDst => glow::ONE_MINUS_DST_COLOR,
+            Self::DstAlpha => glow::DST_ALPHA,
+            Self::OneMinusDstAlpha => glow::ONE_MINUS_DST_ALPHA,
+            Self::SrcAlphaSaturated => glow::SRC_ALPHA_SATURATE,
+            Self::Constant => glow::CONSTANT_ALPHA,
+            Self::OneMinusConstant => glow::ONE_MINUS_CONSTANT_ALPHA,
+        }
+    }
+}
+
+impl crate::BlendOperation {
+    fn to_gles(self) -> u32 {
+        match self {
+            Self::Add => glow::FUNC_ADD,
+            Self::Subtract => glow::FUNC_SUBTRACT,
+            Self::ReverseSubtract => glow::FUNC_REVERSE_SUBTRACT,
+            Self::Min => glow::MIN,
+            Self::Max => glow::MAX,
         }
     }
 }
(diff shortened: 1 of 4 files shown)


blade-graphics/src/gles/command.rs
```diff
@@ -458,6 +468,38 @@ impl crate::VertexFormat {
     }
 }
 
+impl crate::BlendFactor {
+    fn describe(self) -> u32 {
```
kvark, line 472: can we call this method `to_gles`?

blade-graphics/src/gles/command.rs
```diff
@@ -458,6 +468,38 @@ impl crate::VertexFormat {
...
+            Self::OneMinusDstAlpha => glow::ONE_MINUS_DST_ALPHA,
+            Self::SrcAlphaSaturated => glow::SRC_ALPHA_SATURATE,
+            Self::Constant => glow::CONSTANT_ALPHA,
+            Self::OneMinusConstant => glow::ONE_MINUS_CONSTANT_ALPHA,
+        }
+    }
+}
+
+impl crate::BlendOperation {
+    fn describe(self) -> u32 {
```
kvark, line 492: same here

blade-graphics/src/gles/pipeline.rs
```diff
@@ -1,5 +1,6 @@
 use glow::HasContext as _;
 use naga::back::glsl;
+use crate::ColorTargetState;
```
kvark, line 3: let's not change the style in which this module (and the library is written) - just do `crate::ColorTargetState` where it needs to be, no need to declare it here

blade-graphics/src/gles/command.rs
```diff
@@ -396,6 +398,10 @@ impl crate::traits::RenderPipelineEncoder for super::PipelineEncoder<'_> {
         start_instance: u32,
         instance_count: u32,
     ) {
+        for (c, i) in self.color_targets.iter().zip(0..) {
```
kvark, line 401: We don't want to set this on `draw()`. Instead, the blend states shoud be set when the pipeline is switched

blade-graphics/src/gles/command.rs
```diff
@@ -234,6 +235,7 @@ impl super::PassEncoder<'_, super::RenderPipeline> {
             topology: map_primitive_topology(pipeline.topology),
             limits: self.limits,
             vertex_attributes: &pipeline.inner.vertex_attribute_infos,
+            color_targets: &pipeline.inner.color_targets,
```
kvark, line 238: instead of adding the field to `super::PipelineEncoder`, we should use it to issue the relevant commands right here

kvark (changes requested):
Thank you! A few changes are needed

blade-graphics/src/gles/command.rs
lukaslihotzki: This is named like `VertexFormat::describe` defined above, but `to_gles` seems to be the better name. Should the function above also be renamed to `VertexFormat::to_gles`?

kvark: `VertexFormat` is a bit different since it actually describes a format by producing a bunch of characteristics (which are GLES specific), it's not a 1:1 translation like all the rest. I don't have a strong opinion about how `VertexFormat::describe` should be named, leaving up to you.

blade-graphics/src/gles/command.rs
```diff
@@ -737,6 +775,30 @@ impl super::Command {
...
+                        blend_state.color.dst_factor.to_gles(),
+                        blend_state.alpha.src_factor.to_gles(),
+                        blend_state.alpha.dst_factor.to_gles(),
+                    );
+                    gl.blend_equation_draw_buffer(i, blend_state.color.operation.to_gles());
+                } else {
+                    gl.disable(glow::BLEND);
+                }
+            }
+            Self::SetWriteMask(i, write_mask) => {
```
kvark, line 793: There is another problem here - [glColorMaski](https://registry.khronos.org/OpenGL-Refpages/es3/html/glColorMask.xhtml) is only supported on GLES-3.2 (or with "EXT_draw_buffers_indexed").
The GLES backend in Blade should work on WebGL2, which is almost plain GLES-3.0, so if we just merge this all WebGL2 stuff will be broken immediately.

Here is what we could do. At pipeline creation, we'd detect if the color masks are the same for all buffers, or different. Same for the Blend functions. If it's not the same, and we don't support it (i.e. we are in GLES-3.1 or below, and we don't have this extension), we could at least panic and tell the user that GLES can't do this pipeline.
Then, when it's time to set the state, we'd have a condition somewhere that makes it set the state on all of the draw buffers at once.

kvark (changes requested):

blade-graphics/src/gles/mod.rs
```diff
@@ -100,6 +101,8 @@ struct PipelineInner {
     program: glow::Program,
     group_mappings: Box<[ShaderDataMapping]>,
     vertex_attribute_infos: Box<[VertexAttributeInfo]>,
+    blend_states: Box<[Option<crate::BlendState>]>,
+    write_masks: Box<[crate::ColorWrites]>,
```
kvark, line 105: These two always go together in the pipeline up to the point of sending commands to GL. I.e. they are assigned together, they have the same number of elements, and they are processed together.
Can we just merge them into a single array here? We'd keep `enum Command` as you wrote it, but at the pipeline level I think these should be merged

kvark (commented):
Thank you for updating it! One last thing

blade-graphics/src/gles/mod.rs
lukaslihotzki: These boxes do not have the same size if one parameter is the same for all draw buffers and the other is not. Then, the commands also are emitted differently, one simple call for one parameter and multiple PerDrawBuffer-commands for the other parameter.

kvark: Yes, but we don't really benefit from this path ever going to be different between the blend states and write masks. We are only doing this for compatibility reasons (not performance or anything). And for compatibility it's both or nothing.

blade-graphics/src/gles/command.rs
```diff
@@ -227,6 +227,24 @@ impl super::PassEncoder<'_, super::RenderPipeline> {
...
+
+        match &pipeline.inner.blend_states[..] {
+            &[single] => self.commands.push(super::Command::SetBlendState(single)),
+            separate => self
+                .commands
+                .extend(separate.iter().zip(0..).map(|(&blend_state, i)| {
+                    super::Command::SetBlendStatePerDrawBuffer(i, blend_state)
+                })),
+        }
+        match &pipeline.inner.write_masks[..] {
```
kvark, line 239: this could also look less fancy if we treated them together, something like:
```rust
if pipeline.inner.target_states.len() == 1 {
  let state = &pipeline.inner.target_states[0];
   self.commands.push(super::Command::SetBlendState(state.blend));
   self.commands.push(super::Command::SetWriteMask(state.write_mask));
} else {
  for state in pipeline.inner.target_states.iter() {
    ...
  }
}
```
We could also just merge the `Command` variant itself, since it would be used together all the time, further simplifying the whole thing

lukaslihotzki:
I have merged the blend states and write masks according to your preference.

kvark (approved):
Great, thank you!

lukaslihotzki:
Thanks also for your work on supporting GLES! Can you, please, make a release? Then, Zed could be built with `--cfg gles` after a version bump.

PR #127, opened 2024-06-08
Vulkan: Delete smaller descriptor pools.

skejeton:
This fixes a bug where the smaller descriptor sub pools are left unused.

### DIFF
--- a/blade-graphics/src/vulkan/descriptor.rs
+++ b/blade-graphics/src/vulkan/descriptor.rs
@@ -7,6 +7,7 @@ const COUNT_BASE: u32 = 16;
 #[derive(Debug)]
 pub struct DescriptorPool {
     sub_pools: Vec<vk::DescriptorPool>,
+    current_pool: usize,
 }
 
 impl super::Device {
@@ -60,6 +61,7 @@ impl super::Device {
         let vk_pool = self.create_descriptor_sub_pool(COUNT_BASE);
         DescriptorPool {
             sub_pools: vec![vk_pool],
+            current_pool: 0,
         }
     }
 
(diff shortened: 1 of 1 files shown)


kvark:
It wasn't actually a bug, it was done intentionally. The code is simpler if we only use one pool, and with pools growing in size 16x, the extra capacity is just 1/16th, it's arguably negligible.

skejeton:
> It wasn't actually a bug, it was done intentionally. The code is simpler if we only use one pool, and with pools growing in size 16x, the extra capacity is just 1/16th, it's arguably negligible.

Hmm got it, would it be better to release the unused pools then?

kvark:
that would make sense, yeah!

skejeton:
Alright, I'll try doing it this evening.

skejeton:
Done.

blade-graphics/src/vulkan/descriptor.rs
```diff
@@ -100,12 +103,17 @@ impl super::Device {
     }
 
     pub(super) fn reset_descriptor_pool(&self, pool: &mut DescriptorPool) {
-        for &vk_pool in pool.sub_pools.iter() {
+        for &vk_pool in pool.sub_pools[1..].iter() {
```
kvark, line 106: can we do `for vk_pool in pool.sub_pools.drain(1..) {..}` instead?

blade-graphics/src/vulkan/descriptor.rs
```diff
@@ -7,6 +7,7 @@ const COUNT_BASE: u32 = 16;
 #[derive(Debug)]
 pub struct DescriptorPool {
     sub_pools: Vec<vk::DescriptorPool>,
+    stage: usize,
```
kvark, line 10: let's call it something that's not confused with "shader stage"
maybe `growth_iter`?

kvark (commented):
Thank you!

skejeton:
fixed

kvark (approved):

PR #128, opened 2024-06-10
Remove unwraps from init()

ConradIrwin:
This will allow us to present a sensible error message to users when we cannot
open a window.

Co-Authored-By: Mikayla <mikayla@zed.dev>


mikayla-maki:
Consider this a draft PR, as we need to make a similar change with other Backends.

kvark (approved):

mikayla-maki:
Assuming CI passes, this PR should be ready for review :)

kvark (approved):

Issue #117, opened 2024-05-14
Should `descriptor_pool_info` include `FREE_DESCRIPTOR_SET` flag?


dannycoates:
From https://github.com/zed-industries/zed/issues/10018#issuecomment-2107354282

progmboy:
After several hours of debugging, I can confirm that this issue only occurs with Intel's drivers.

```
filename: igvk64.dll
fileversion: 31.0.101.4502
filedescription: Vulkan(R) Driver for Intel(R) Graphics Accelerator
```

```c
__int64 __fastcall vkDescriptorAllocatePoolSet(__int64 a1, __int64 a2, __int64 a3)
{
	...
	// Check descriptor_pool->dword424 is bigger than set.
	// the throw VK_ERROR_OUT_OF_POOL_MEMORY error.
	//
	// When this error occurs.
	// 0:000> dd 0x00000212bdec4878+424 l1
	// 00000212`bdec4c9c  0000ea60
	// 0:000> dd 0x00000212bdec4878+420 l1
	// 00000212`bdec4c98  0000ea60
	//
	if ( v11 && (unsigned int)(descriptor_pool->dword424 + 1) > descriptor_pool->dword420 )
        {
            sub_180470550(&v36, 20LL);
            v37[0] = v36;
            CxxThrowException(v37, (_ThrowInfo *)&_TI1_AUException_VK__);
        }
	...

	// here inc dword424 
        *(_DWORD *)(v14 + 68) = descriptor_pool->dword424++;
        v30 = v12;
        v20 = v11;
        v21 = playout;
}
```

```c
__int64 __fastcall vkResetDescriptorPool(__int64 hDevice, struct_descriptor_pool *descriptor_pool)
{
	// 
	// VK_DESCRIPTOR_POOL_CREATE_FREE_DESCRIPTOR_SET_BIT = 1
	// defalut descriptor_pool->flags == 0
	// 
	if ( descriptor_pool->flags ){
		...
	
		//
		//  void __fastcall sub_18039C3B0(struct_descriptor_pool *descriptor_pool, __int64 a2)
		//  {
		//	 ...
		//	 // ONLY this way to dec descriptor_pool->dword424
		//   descriptor_pool->dword424 -= *(_DWORD *)(a2 + 68) != -1;
  		//   v16 = *(unsigned int *)(a2 + 68);
		//   ...
		// 	}
		//
	
		sub_18039C3B0(descriptor_pool, v11);
		...
	}else{
[...]

PR #129, opened 2024-06-11
vk: add workaround for Intel HD/UHD memory leak (#117)

skejeton:
This will detect an Intel graphics card and apply the fix for leaky descriptor pools.

### DIFF
--- a/blade-graphics/src/vulkan/descriptor.rs
+++ b/blade-graphics/src/vulkan/descriptor.rs
@@ -45,18 +45,12 @@ impl super::Device {
             max_inline_uniform_block_bindings: max_sets,
             ..Default::default()
         };
-        let descriptor_pool_info = if self.workarounds.intel_fix_descriptor_pool_leak {
-            vk::DescriptorPoolCreateInfo::default()
-                .max_sets(max_sets)
-                .flags(vk::DescriptorPoolCreateFlags::FREE_DESCRIPTOR_SET)
-                .pool_sizes(&descriptor_sizes)
-                .push_next(&mut inline_uniform_block_info)
-        } else {
-            vk::DescriptorPoolCreateInfo::default()
-                .max_sets(max_sets)
-                .pool_sizes(&descriptor_sizes)
-                .push_next(&mut inline_uniform_block_info)
-        };
+
+        let descriptor_pool_info = vk::DescriptorPoolCreateInfo::default()
+            .max_sets(max_sets)
+            .flags(self.workarounds.extra_descriptor_pool_create_flags)
+            .pool_sizes(&descriptor_sizes)
+            .push_next(&mut inline_uniform_block_info);
 
         unsafe {
             self.core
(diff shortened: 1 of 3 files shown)


blade-graphics/src/vulkan/init.rs
```diff
@@ -143,6 +150,9 @@ unsafe fn inspect_adapter(
     let name = ffi::CStr::from_ptr(properties.device_name.as_ptr());
     log::info!("Adapter {:?}", name);
 
+    let vendor_id = properties2_khr.properties.vendor_id;
+    let intel_fix_descriptor_pool_leak = vendor_id == db::intel::VENDOR;
```
kvark, line 154: can we limit to Windows only?

blade-graphics/src/vulkan/init.rs
```diff
@@ -143,6 +150,9 @@ unsafe fn inspect_adapter(
     let name = ffi::CStr::from_ptr(properties.device_name.as_ptr());
     log::info!("Adapter {:?}", name);
 
+    let vendor_id = properties2_khr.properties.vendor_id;
```
kvark, line 153: if we are doing this, let's also use `vendor_id` when computing `intel_unable_to_present` consistently

blade-graphics/src/vulkan/mod.rs
```diff
@@ -23,6 +23,7 @@ struct RayTracingDevice {
 struct Workarounds {
     extra_sync_src_access: vk::AccessFlags,
     extra_sync_dst_access: vk::AccessFlags,
+    intel_fix_descriptor_pool_leak: bool,
```
kvark, line 26: Perhaps we can instead store it as
```
extra_descriptor_pool_create_flags: vk::DescriptorPoolCreateFlags
```
Or at least call it something neutral, like `force_free_descriptor_set_flag`

blade-graphics/src/vulkan/descriptor.rs
```diff
@@ -45,10 +45,19 @@ impl super::Device {
...
             ..Default::default()
         };
-        let descriptor_pool_info = vk::DescriptorPoolCreateInfo::default()
-            .max_sets(max_sets)
-            .pool_sizes(&descriptor_sizes)
-            .push_next(&mut inline_uniform_block_info);
+        let descriptor_pool_info = if self.workarounds.intel_fix_descriptor_pool_leak {
+            vk::DescriptorPoolCreateInfo::default()
+                .max_sets(max_sets)
+                .flags(vk::DescriptorPoolCreateFlags::FREE_DESCRIPTOR_SET)
```
kvark, line 51: this if/else should be moved inside the `flags`
Either
```rust
.flags(self.workarounds.extra_descriptor_pool_create_flags)
```
Or
```rust
.flags(if ... {} else {})
```

kvark (commented):
Thank you for the PR!
A few small changes are needed, but overall the approach is good

skejeton:
fixed

kvark (approved):
great, thank you!

PR #138, opened 2024-07-12
Use builtin transpose and map instead of if let

WeetHet:
This just looks a bit nicer in my opinion


kvark:
Let's keep the if/let - there is less sugar involved.

Issue #142, opened 2024-07-17
Allow querying the name and driver info of the chosen adapter


jansol:
Corresponding to GL_RENDERER and GL_DEVICE_NAME and similar on Vulkan.

These would be very useful to have as part of Zed's system info.

PR #143, opened 2024-07-18
Expose device_information

ConradIrwin:
We'd like to be able to tell users not to use a software renderer, and to provide more information in the case of failures.

Fixes: kvark/blade#142

I have not yet implemented this on metal, but can do so if this approach seems sound to you.

### DIFF
--- a/blade-graphics/src/metal/mod.rs
+++ b/blade-graphics/src/metal/mod.rs
@@ -465,6 +465,10 @@ impl Context {
         }
     }
 
+    pub fn device_information(&self) -> &crate::DeviceInformation {
+        todo!();
+    }
+
     /// Get the CALayerMetal for this surface, if any.
     /// This is platform specific API.
     pub fn metal_layer(&self) -> Option<metal::MetalLayer> {
--- a/blade-graphics/src/metal/mod.rs
+++ b/blade-graphics/src/metal/mod.rs
@@ -49,6 +49,7 @@ pub struct Context {
     surface: Option<Mutex<Surface>>,
     capture: Option<metal::CaptureManager>,
     info: DeviceInfo,
+    device_information: crate::DeviceInformation,
 }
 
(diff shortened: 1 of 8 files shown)


kvark (commented):
Generally speaking, exposing some information about the device is good and helpful.

However, I would very much like to squeeze this information to a form that is useful, as opposed to blindly exposing everything that Vulkan gives us. For example, Vulkan drivers aren't even in agreement about what is considered an integrated GPU and what not. This line is being blurred, and it's not really actionable by the user anyway.

What the user does need to know is if the adapter is emulated in software. That is definitely a strong distinction, where the line isn't blurred. So, we would ideally expose something like this:
```
struct DeviceInformation {
  name: String,
  is_software_emulated: bool,
}
```

ConradIrwin:
@kvark :+1: can make that change.

I had a few people on the linux channel suggest it would be useful to return the driver info too to help people debug what's happening. Is that information obtainable a different way, or should I keep returning that?

ConradIrwin:
@kvark updated to just return a bool. Kept all three strings, but can simplify if you'd prefer

blade-graphics/src/metal/mod.rs
```diff
@@ -465,6 +465,10 @@ impl Context {
         }
     }
 
+    pub fn device_information(&self) -> &crate::DeviceInformation {
+        todo!();
```
kvark, line 469: Can we return a default implementation instead? Crashing isn't fun for this

kvark (commented):
Sorry, one last thing

ConradIrwin:
@kvark I'm keen to ship a new build to zed preview on Wednesday. I can pull this in, but is #144 something you feel comfortable with too?

kvark:
@ConradIrwin I don't want to merge #144 until we confirm that it helps people.

PR #144, opened 2024-07-18
Try to actually acquire a frame on Intel/Linux upon adapter inspection

kvark:
Another take at the problem previously addressed by #94/#95/#97

### DIFF
--- a/blade-graphics/src/vulkan/init.rs
+++ b/blade-graphics/src/vulkan/init.rs
@@ -1,6 +1,6 @@
 use ash::{amd, ext, khr, vk};
 use naga::back::spv;
-use std::{ffi, fs, mem, sync::Mutex};
+use std::{ffi, mem, sync::Mutex};
 
 use crate::NotSupportedError;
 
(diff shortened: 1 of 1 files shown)


ConradIrwin:
@kvark I'm planning to pull this into Zed so we can get feedback from people.

kvark:
Looks like this path faced some issues for end users

apricotbucket28:
Hey, could this be rebased on `master` to fix #168? Thanks!

kvark:
@apricotbucket28 done

gretchenfrage:
Some information that may be relevant to this, especially if kvark finds it inconvenient to test Wayland locally:

- I still experience the issue described in https://github.com/zed-industries/zed/issues/17005 even if I switch Zed to use this branch.
- On that issue, I just added a backtrace of a segfault I captured with gdb, of encountering this issue. This may be useful as I didn't see a segfault already posted on related issue threads.

kvark:
This path is no longer viable, since as of #203 we don't have a specific surface any more when creating a GPU context.
We'll need to find a better path.

PR #148, opened 2024-07-21
Multiple backends

lukaslihotzki:
This draft shows an approach how both Vulkan and GLES can be supported by a single executable. In the first commit, all traits specific to the backend are moved from the crate root to the common directory, which is included in both backends as separate modules using `path` attributes. In the following commits, the traits are made more abstract from specific backends:

1. The dependency of HasShaderBinding and ShaderBindable is inverted, so now, ShaderBindable depends on HasShaderBinding. Practically, none or both traits are implemented, so it makes no difference. But with this change, HasShaderBinding can be implemented independently of the backend.
2. The traits ResourceDevice and CommandDevice are made independent from the backend by making TextureViewDesc generic on the inner Texture.
3. Buffer handling becomes generic by introducing the Buffer trait. This allows to make the belt in blade-util generic over multiple backends.

I think these changes should be sufficient to support multiple backends in Zed, for example. However, making blade even more generic with extra traits (for example, Texture handling) would probably make app development easier.


kvark:
Hi @lukaslihotzki and thank you for experimenting with this!

The graphics library has a fairly large API. Making it polymorphic across backends is taking a tall at the code clarity and quality.

I wonder if we can expose it such that the user would enable "gles" as a feature and see this API:
```rust
blade_graphics::Device; // main device
blade_graphics::fallback::Device; // GLES device
```

So essentially GLES feature would bring another implementation exposed inside "fallback" namespace, and the user can use both. That would allow Zed (and other users) to potentially have different renderers, e.g. `BladeRenderer` and `BladeFallbackRenderer`, thus shifting the polymorphism to a higher level.

lukaslihotzki:
With #150 and #151, the diff shrinked from `+434 −352` to `+321 −254`, so some progress is made on this. Nice!

theofficialgman:
Any interest in picking this back up? This is a blocker for zed on many devices (already mentioned in the issue reports related to this).

PR #150, opened 2024-07-24
Make ResourceDevice actually generic over the backend

lukaslihotzki:
The ResourceDevice trait is intended not to depend on the specific backend. This is shown by the type parameters that are used in the function signatures (such as `Self::Texture`) instead of `super::Texture`. However, `super::TextureViewDesc` contains `super::Texture`, so the trait indirectly depends on the concrete `super::Texture`, but not on all the other type parameters. Therefore, this seems like an oversight.

To fix this inconsistency, this PR makes `crate::TextureViewDesc` generic over the contained Texture, so the trait can use this struct with the correct `Self::Texture` type parameter. Advantage of this solution is that users don't need to change the code if they just pass the TextureViewDesc to `create_texture_view`, even if there were different implementations of `ResourceDevice`. Alternative solution would be to exclude the actual `Texture` from `TextureViewDesc` and just specify it as first parameter in `create_texture_view`. Another solution would be to add a `Self::TextureViewDesc` type parameter to `ResourceDevice`, but then, it would be difficult for users to create the `TextureViewDesc` in a generic way.

I would be happy if this inconsistency is fixed with any solution (the mentioned ones or another).

### DIFF
--- a/blade-graphics/src/lib.rs
+++ b/blade-graphics/src/lib.rs
@@ -388,9 +388,9 @@ pub struct TextureSubresources {
 }
 
 #[derive(Debug)]
-pub struct TextureViewDesc<'a> {
+pub struct TextureViewDesc<'a, T> {
     pub name: &'a str,
-    pub texture: Texture,
+    pub texture: T,
     pub format: TextureFormat,
     pub dimension: ViewDimension,
     pub subresources: &'a TextureSubresources,
--- a/blade-graphics/src/lib.rs
+++ b/blade-graphics/src/lib.rs
@@ -388,9 +388,8 @@ pub struct TextureSubresources {
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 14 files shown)


lukaslihotzki:
This change is related to #148, but it makes sense on its own.

blade-graphics/src/lib.rs
```diff
@@ -388,9 +388,9 @@ pub struct TextureSubresources {
 }
 
 #[derive(Debug)]
-pub struct TextureViewDesc<'a> {
+pub struct TextureViewDesc<'a, T> {
```
kvark, line 391: I think we are doing this we should instead just change the signature `create_texture_view` to get the texture as the first argument and the descriptor as the second. This way, the descriptor isn't going to be generic.

kvark (approved):
Thank you!

PR #152, opened 2024-07-29
Add support for FreeBSD

180watt:
Just a few cfg changes to allow Vulkan rendering on FreeBSD. Tested both GLES and Vulkan to be working on FreeBSD-current with drm-5.15.

### DIFF
--- a/blade-graphics/Cargo.toml
+++ b/blade-graphics/Cargo.toml
@@ -26,7 +26,7 @@ metal = "0.29"
 objc = "0.2.5"
 naga = { workspace = true, features = ["msl-out"] }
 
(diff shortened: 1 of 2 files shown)


kvark (approved):

180watt:
What should I follow from the CI suggestions? I could reformat some of the cfg lines if you'd like, but I'm unsure if the Windows build error is mine.

180watt:
okay, just pushed after a `cargo fmt`

180watt:
nothing like forgetting to `git add`

180watt:
apologizes for fiddling. just trying to get auto-merge to be happy.

PR #153, opened 2024-08-04
vk: ignore non-coherent host-visible memory types

kvark:
Relevant to https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/30499
Since Blade doesn't expose API to do explicit memory flush/invalidation, let's ignore the affected memory types.


Jaakkonen:
Tested this one and the flickering seems to go away! I couldn't reproduce the flickering that reliably on the Blade's own examples but changed Zed editor to use this branch and against that this fixes the issue.

PR #162, opened 2024-08-26
Correctly drop Vulkan objects

MatinAniss:
There are some Vulkan objects that are created when creating the context that do not implement drop semantics and require them to be explicitly destroyed. This PR correctly destroys these objects.

This should fix/help this issue observed in Zed https://github.com/zed-industries/zed/issues/13346, where memory leaks due to creating and destroying the context's by creating and destroying windows in Zed.

This needed explicit object destruction is documented here [create_instance](https://docs.rs/ash/latest/ash/struct.Entry.html#method.create_instance), [create_device](https://docs.rs/ash/latest/ash/struct.Instance.html#method.create_device), and [create_surface](https://docs.rs/ash-window/latest/ash_window/fn.create_surface.html).

### DIFF
--- a/blade-graphics/src/vulkan/init.rs
+++ b/blade-graphics/src/vulkan/init.rs
@@ -1015,3 +1015,17 @@ impl super::Context {
         }
     }
 }
+
+impl Drop for super::Context {
+    fn drop(&mut self) {
+        unsafe {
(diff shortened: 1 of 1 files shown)


kvark (approved):

kvark:
Thank you!

PR #167, opened 2024-08-28
Report an error on initialization failure

ConradIrwin:
About 15% of our graphics issues are a failure to initialize. I assume it's
this unwrap (though I have no proof)


kvark (approved):
It's very weird to see this one failing. We are checking for all the features and extensions before choosing a device, so this must be something easily uncovered with vulkan validation layers. If we knew what kind of machine it was failing on, we could grab a system profile for the VVL to emulate, and reproduce the error.

PR #168, opened 2024-08-30
Destroy queue before device

apricotbucket28:
See: https://github.com/zed-industries/zed/issues/17005

Right now, the device is being destroyed before the queue. This causes crashes at least on Wayland (both in Zed and the examples in this repo).

There are still some Vulkan validation errors related to objects not being destroyed, but it's not crashing anymore at least 😅


kvark (approved):
this isn't actually destroying the queue before anything
queue isn't being destroyed in Vulkan
but it fixes the order of destruction for the queue semaphores

PR #193, opened 2024-10-15
Unify temporal and spatial taps

kvark:
Unless ReSTIR is multi-pass (e.g. #192), there is really no difference between spatial taps and the temporal tap. This PR unifies them together, which simplifies the code a bit.


kvark:
Replaced by #194

PR #197, opened 2024-10-31
Update `egui` to `0.29` and `winit` to `0.30`

### LINKED ISSUE #196: Update to egui 0.29
`egui` has now updated to 0.29 would appreciate if blade-egui can do the same and a bumped crate be released on crates.io :)

This also allows to use a newer version of winit, which is super useful (new decorations on Windows 11!)

apekros:
Fixes #196

`winit` in `0.30` dropped `WindowBuilder` in favour of `WindowAttributes`, so this has be fixed in the examples.

`EventLoop::run_app` should probably be used instead of deprecated `EventLoop::run` in examples too


kvark (approved):
Nice work!

PR #199, opened 2024-11-07
egl: destroy the previous surface on resize

kvark:
Closes #5505
(I hope... please re-open otherwise!)
Thanks @andreymal for the fix


andreymal:
If I understand correctly, you called `destroy_surfaces` after `create_platform_window_surface` which probably means it will still crash

PR #206, opened 2024-11-21
Refactor examples to remove use of deprecated winit functions

apekros:
Just a quick draft, as I was not sure if you would be picky about the structure in the examples. I've just refactored the existing `Example` to be `ExampleInner` which is now contained within a new `Example` struct which holds the window/inner. I did this to minimise the amount of changes. 

I can however go for a different approach and keep the Example struct, but instantiate everything within new() that does not require a Window, and then have a secondary `instantiate` (no good name comes to mind right now), that can be called with a window on `ApplicationHandler::resumed()`

Will await your thoughts before applying this across all the examples

### DIFF
--- a/examples/ray-query/main.rs
+++ b/examples/ray-query/main.rs
@@ -2,6 +2,7 @@
 
 use blade_graphics as gpu;
 use std::{mem, ptr, time};
+use winit::application::ApplicationHandler;
 
 const TORUS_RADIUS: f32 = 3.0;
 const TARGET_FORMAT: gpu::TextureFormat = gpu::TextureFormat::Rgba16Float;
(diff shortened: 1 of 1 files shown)


kvark:
Ideally, the `struct Example` doesn't care about the window, yeah. If you can prototype the (more invasive) alternative that would be awesome!

apekros:
> Ideally, the `struct Example` doesn't care about the window, yeah. If you can prototype the (more invasive) alternative that would be awesome!

I'll look into if I can do it without the window being contained within `Example` but looks like that's what the winit team envisions it as being in their [examples](https://github.com/rust-windowing/winit/blob/master/examples/run_on_demand.rs).

PR #203, opened 2024-11-15
Surface API

kvark:
Closes #82
TODO:
- [x] API
- [x] Vulkan
- [x] Metal
- [x] GLES
- [ ] Update in Zed
  - I'm confident in the direction here, so testing on Zed can be done in follow-up

Begone the `init_windowed` method. There is only one `init` now. The other one is replaced by `create_surface`.

PR #213, opened 2024-12-04
Support for Multisampled Anti-Aliasing

### LINKED ISSUE #198: Support for multisampled textures + resolve?\
I see there are some indications in the code that `sample_count` was almost added to the texture descriptor but don't see it implemented yet.

What would be the challenges in supporting multisampled textures and resolve targets for render passes? I would be happy to help though I have mostly experience with wgpu:s user-facing code

(side note: Such a fresh breath of air to read this code! It's all so simple.. blade-graphics looks very promising and I would like to give blade a try for our rendering engine at my company that's currently using wgpu (for this game https://idno.se/swap/). Seeing your presentation and descirbing the simple binding model in blade really motivated me after having to deal with so much pipeline description and setup of bind groups for our multiple shader permutations... )

EriKWDev:
closes #198 

### Changes
This PR introduces a `sample_count: u32` to `TextureDesc` as well as a `multisample_state: MultisampleState` to the `RenderPipelineDesc`. Also added `alpha_to_coverage` while at it.

```rust
#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
pub struct MultisampleState {
    pub sample_count: u32,
    pub sample_mask: u64,
    pub alpha_to_coverage: bool,
}

impl Default for MultisampleState {
    fn default() -> Self {
        Self {
            sample_count: 1,
            sample_mask: !0,
            alpha_to_coverage: false,
        }
    }
}
```

Together with the existing `FinishOp::ResolveTo(TextureView)`, a multisampled renderpass can now be described and executed

### Implementations
<details>
  <summary>Vulkan</summary>

The rendering attachment needs to be told to resolve
```rust
if let crate::FinishOp::ResolveTo(resolve_view) = rt.finish_op {
    vk_info = vk_info
        .resolve_image_view(resolve_view.raw)
        .resolve_image_layout(vk::ImageLayout::GENERAL)
        .resolve_mode(vk::ResolveModeFlags::AVERAGE);
}

```

The store_op is currently always set to `DONT_CARE` as in many cases for msaa rendering the resolved texture is the only one required. Would be nice to be able to specify this behaviour in blade as well though: 

```rust
vk_info.store_op = match rt.finish_op {
    crate::FinishOp::ResolveTo(..) => {
        /*
            TODO: DONT_CARE is most optimal in many cases where the msaa texture itself is never read afterwards but only the resolved,
                  but how can the user specify this in blade?
                  https://docs.vulkan.org/samples/latest/samples/performance/msaa/README.html#_best_practice_summary
        */

        // vk::AttachmentStoreOp::STORE
        vk::AttachmentStoreOp::DONT_CARE
    }
};

```

For the texture creation, a `SampleCountFlags` is constructed for the `vk::ImageCreateInfo`:
```rust
/// in vulkan/resource.rs:268
samples: vk::SampleCountFlags::from_raw(desc.sample_count),
```

and for the pipeline, I looked at how wgpu does it and came to the following code

```rust
let vk_sample_mask = [
    desc.multisample_state.sample_mask as u32,
    (desc.multisample_state.sample_mask >> 32) as u32,
];

let vk_multisample = vk::PipelineMultisampleStateCreateInfo::default()
    .rasterization_samples(vk::SampleCountFlags::from_raw(
        desc.multisample_state.sample_count,
    ))
    .alpha_to_coverage_enable(desc.multisample_state.alpha_to_coverage)
    .sample_mask(&vk_sample_mask);

```
  
</details>

<details>
  <summary>Metal</summary>
The metal backend already had support for the `ResolveTo` finishop so no changes needed for the renderpipeline. The creation of the renderpipeline required description though:

```rust
descriptor.set_raster_sample_count(desc.multisample_state.sample_count);
descriptor.set_alpha_to_coverage_enabled(desc.multisample_state.alpha_to_coverage);
```

For the texture, only specifying the sample_count was required

```rust
/// in metal/resource.rs:205
descriptor.set_sample_count(desc.sample_count as u64);
```
</details>

<details>
  <summary>OpenGL ES</summary>

This was the hardest one as I am very unfamiliar with msaa in opengl, but in summary the texture needs to be created with `renderbuffer_storage_multisample` if it is a rendertarget and with a `glow::TEXTURE_2D_MULTISAMPLE` target type if it is a normal texture.

The sample count is then set with `tex_storage_2d_multisample`, but there is now no-longer a way to specify `mip_count` so don't know if that is even possible.
```rust
/*
    TODO(ErikWDev): How to set mip count and sample count? Not possible in gles?
*/
gl.tex_storage_2d_multisample(
    target,
    desc.sample_count as i32,
    format_desc.internal,
    desc.size.width as i32,
    desc.size.height as i32,
    true,
);
```

For the rendering, I use `blit_framebuffer` to blit the msaa texture onto the resolve target which required turning the renderbuffers into FBO:s. Currently, the FBO:s are created and deleted ad-hoc but could be saved as done in wgpu-hal. See #198

```rust
/// in gles/command.rs:814
Self::BlitFramebuffer { from, to } => {
    /*
        TODO(ErikWDev): Validate
    */

    let target_from = match from.inner {
        super::TextureInner::Renderbuffer { raw } => raw,
        _ => panic!("Unsupported resolve between non-renderbuffers"),
    };
    let target_to = match to.inner {
        super::TextureInner::Renderbuffer { raw } => raw,
        _ => panic!("Unsupported resolve between non-renderbuffers"),
    };

    let framebuf_from = gl.create_framebuffer().unwrap();
    let framebuf_to = gl.create_framebuffer().unwrap();

    gl.bind_framebuffer(glow::READ_FRAMEBUFFER, Some(framebuf_from));
    gl.framebuffer_renderbuffer(
        glow::READ_FRAMEBUFFER,
        glow::COLOR_ATTACHMENT0, // NOTE: Assuming color attachment
        glow::RENDERBUFFER,
        Some(target_from),
    );

    gl.bind_framebuffer(glow::DRAW_FRAMEBUFFER, Some(framebuf_to));
    gl.framebuffer_renderbuffer(
        glow::DRAW_FRAMEBUFFER,
        glow::COLOR_ATTACHMENT0, // NOTE: Assuming color attachment
        glow::RENDERBUFFER,
        Some(target_to),
    );
    assert_eq!(
        gl.check_framebuffer_status(glow::DRAW_FRAMEBUFFER),
        glow::FRAMEBUFFER_COMPLETE,
        "DRAW_FRAMEBUFFER is not complete"
    );

    gl.blit_framebuffer(
        0,
        0,
        from.target_size[0] as _,
        from.target_size[1] as _,
        0,
        0,
        to.target_size[0] as _,
        to.target_size[1] as _,
        glow::COLOR_BUFFER_BIT, // NOTE: Assuming color
        glow::NEAREST,
    );

    gl.bind_framebuffer(glow::READ_FRAMEBUFFER, None);
    gl.bind_framebuffer(glow::DRAW_FRAMEBUFFER, None);

    gl.delete_framebuffer(framebuf_from);
    gl.delete_framebuffer(framebuf_to);
}


``` 

</details>

### Testing
The vulkan implementation has been validated on an AMD RX 6600 on debian, an integrated intel GPU on Fedora as well as a GTX 1050 on Windows and it all seems to work (the particle sample). After inspection in renderdoc the result is as expected.

The metal implementation has been tested on a Mac Mini M4 and works. However, particle example seems to break after the sample count is adjusted in runtime for some reason.

The Opengl ES implementation has **not been tested**!!!

### Particle Example
I changed the particle example to now utilize msaa which requires it to keep a msaa texture with the desired `sample_count` available and recreated upon resizing as well as a FinishOp::ResolveTo to resolve the msaa texture to the acquired frame texture as such:

```rust
if let mut pass = self.command_encoder.render(
    "draw",
    gpu::RenderTargetSet {
        colors: &[gpu::RenderTarget {
            view: self.msaa_view,
            init_op: gpu::InitOp::Clear(gpu::TextureColor::OpaqueBlack),
            finish_op: gpu::FinishOp::ResolveTo(frame.texture_view()),
        }],
        depth_stencil: None,
    },
) {
    self.particle_system
        .draw(&mut pass, screen_desc.physical_size);
    self.gui_painter
        .paint(&mut pass, gui_primitives, screen_desc, &self.context);
}
``` 

### Egui
Furthermore, the `blade_egui` renderer also requires information about the multisample state since it has a pipeline that now needs information about the texture it's going to render to so it's initializer now takes a sample_count as well:

```rust
pub fn new(
    info: blade_graphics::SurfaceInfo,
    context: &blade_graphics::Context,
    sample_count: u32,
) -> GuiPainter {
// ...
}
```

---

Let me know what needs changing or if testing fails somewhere. Especially the Opengl ES implementation as I didn't know how to run using it

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,8 +28,8 @@ profiling = "1"
 slab = "0.4"
 strum = { version = "0.25", features = ["derive"] }
 web-sys = "0.3.60"
-winit = { version = "0.30" }
-# winit = { version = "0.30", default-features=false, features=["x11", "rwh_06"] }
+# winit = { version = "0.30" }
+winit = { version = "0.30", default-features=false, features=["x11", "rwh_06"] }
 
 [lib]
 
@@ -91,7 +91,7 @@ del-geo = "=0.1.29"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 # see https://github.com/emilk/egui/issues/4270
-egui-winit = "0.29"
+egui-winit = { version="0.29", default-features=false, features=["links"] }
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
 # see https://github.com/emilk/egui/issues/4270
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ slab = "0.4"
 strum = { version = "0.25", features = ["derive"] }
 web-sys = "0.3.60"
 winit = { version = "0.30" }
+# winit = { version = "0.30", default-features=false, features=["x11", "rwh_06"] }
 
 [lib]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,6 +94,8 @@ del-geo = "=0.1.29"
 egui-winit = "0.29"
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
+# see https://github.com/emilk/egui/issues/4270
+egui-winit = { version="0.29", default-features=false, features=["links"] }
 console_error_panic_hook = "0.1.7"
 console_log = "1"
 web-sys = { workspace = true, features = ["Window"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,8 +28,7 @@ profiling = "1"
 slab = "0.4"
 strum = { version = "0.25", features = ["derive"] }
 web-sys = "0.3.60"
-# winit = { version = "0.30" }
-winit = { version = "0.30", default-features=false, features=["x11", "rwh_06"] }
+winit = { version = "0.30" }
 
 [lib]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ profiling = "1"
 slab = "0.4"
 strum = { version = "0.25", features = ["derive"] }
 web-sys = "0.3.60"
-winit = "0.30"
+winit = { version = "0.30" }
 
 [lib]
 
--- a/blade-egui/src/lib.rs
+++ b/blade-egui/src/lib.rs
@@ -129,7 +129,11 @@ impl GuiPainter {
     /// It supports renderpasses with only a color attachment,
     /// and this attachment format must be The `output_format`.
     #[profiling::function]
-    pub fn new(info: blade_graphics::SurfaceInfo, context: &blade_graphics::Context) -> Self {
+    pub fn new(
+        info: blade_graphics::SurfaceInfo,
+        context: &blade_graphics::Context,
+        sample_count: u32,
+    ) -> Self {
         let shader = context.create_shader(blade_graphics::ShaderDesc {
             source: SHADER_SOURCE,
         });
@@ -162,7 +166,10 @@ impl GuiPainter {
                 }),
                 write_mask: blade_graphics::ColorWrites::all(),
             }],
-            multisample_state: blade_graphics::MultisampleState::default(),
+            multisample_state: blade_graphics::MultisampleState {
+                sample_count,
+                ..Default::default()
+            },
         });
 
         let belt = BufferBelt::new(BufferBeltDescriptor {
--- a/blade-egui/src/lib.rs
+++ b/blade-egui/src/lib.rs
@@ -129,11 +129,7 @@ impl GuiPainter {
     /// It supports renderpasses with only a color attachment,
     /// and this attachment format must be The `output_format`.
     #[profiling::function]
-    pub fn new(
-        info: blade_graphics::SurfaceInfo,
-        context: &blade_graphics::Context,
-        sample_count: u32,
-    ) -> Self {
+    pub fn new(info: blade_graphics::SurfaceInfo, context: &blade_graphics::Context) -> Self {
         let shader = context.create_shader(blade_graphics::ShaderDesc {
             source: SHADER_SOURCE,
         });
@@ -166,10 +162,7 @@ impl GuiPainter {
                 }),
                 write_mask: blade_graphics::ColorWrites::all(),
             }],
-            multisample_state: blade_graphics::MultisampleState {
-                sample_count,
-                ..Default::default()
-            },
+            multisample_state: Default::default(),
         });
 
         let belt = BufferBelt::new(BufferBeltDescriptor {
--- a/blade-egui/src/lib.rs
+++ b/blade-egui/src/lib.rs
@@ -71,6 +71,7 @@ impl GuiTexture {
             mip_level_count: 1,
             dimension: blade_graphics::TextureDimension::D2,
             usage: blade_graphics::TextureUsage::COPY | blade_graphics::TextureUsage::RESOURCE,
+            sample_count: 1,
         });
         let view = context.create_texture_view(
             allocation,
@@ -161,6 +162,7 @@ impl GuiPainter {
                 }),
                 write_mask: blade_graphics::ColorWrites::all(),
             }],
+            multisample_state: blade_graphics::MultisampleState::default(),
         });
 
         let belt = BufferBelt::new(BufferBeltDescriptor {
--- a/blade-graphics/src/gles/command.rs
+++ b/blade-graphics/src/gles/command.rs
@@ -185,6 +185,10 @@ impl super::CommandEncoder {
             if let crate::FinishOp::Discard = rt.finish_op {
                 invalidate_attachments.push(attachment);
             }
+            if let crate::FinishOp::ResolveTo(to) = rt.finish_op {
+                self.commands
+                    .push(super::Command::BlitFramebuffer { from: rt.view, to });
+            }
         }
         if let Some(ref rt) = targets.depth_stencil {
             let attachment = match rt.view.aspects {
@@ -806,6 +810,65 @@ impl super::Command {
                     None,
                 );
             }
+
+            Self::BlitFramebuffer { from, to } => {
+                /*
+                    TODO(ErikWDev): Validate
+                */
+
+                let target_from = match from.inner {
+                    super::TextureInner::Renderbuffer { raw } => raw,
+                    _ => panic!("Unsupported resolve between non-renderbuffers"),
+                };
+                let target_to = match to.inner {
+                    super::TextureInner::Renderbuffer { raw } => raw,
+                    _ => panic!("Unsupported resolve between non-renderbuffers"),
+                };
+
+                let framebuf_from = gl.create_framebuffer().unwrap();
+                let framebuf_to = gl.create_framebuffer().unwrap();
+
+                gl.bind_framebuffer(glow::READ_FRAMEBUFFER, Some(framebuf_from));
+                gl.framebuffer_renderbuffer(
+                    glow::READ_FRAMEBUFFER,
+                    glow::COLOR_ATTACHMENT0, // NOTE: Assuming color attachment
+                    glow::RENDERBUFFER,
+                    Some(target_from),
+                );
+
+                gl.bind_framebuffer(glow::DRAW_FRAMEBUFFER, Some(framebuf_to));
+                gl.framebuffer_renderbuffer(
+                    glow::DRAW_FRAMEBUFFER,
+                    glow::COLOR_ATTACHMENT0, // NOTE: Assuming color attachment
+                    glow::RENDERBUFFER,
+                    Some(target_to),
+                );
+                assert_eq!(
+                    gl.check_framebuffer_status(glow::DRAW_FRAMEBUFFER),
+                    glow::FRAMEBUFFER_COMPLETE,
+                    "DRAW_FRAMEBUFFER is not complete"
+                );
+
+                gl.blit_framebuffer(
+                    0,
+                    0,
+                    from.target_size[0] as _,
+                    from.target_size[1] as _,
+                    0,
+                    0,
+                    to.target_size[0] as _,
+                    to.target_size[1] as _,
+                    glow::COLOR_BUFFER_BIT, // NOTE: Assuming color
+                    glow::NEAREST,
+                );
+
+                gl.bind_framebuffer(glow::READ_FRAMEBUFFER, None);
+                gl.bind_framebuffer(glow::DRAW_FRAMEBUFFER, None);
+
+                gl.delete_framebuffer(framebuf_from);
+                gl.delete_framebuffer(framebuf_to);
+            }
+
             Self::BindAttachment {
                 attachment,
                 ref view,
--- a/blade-graphics/src/gles/command.rs
+++ b/blade-graphics/src/gles/command.rs
@@ -813,37 +813,55 @@ impl super::Command {
 
             Self::BlitFramebuffer { from, to } => {
                 /*
-                    TODO(ErikWDev): Validate
+                    TODO: Framebuffers could be re-used instead of being created on the fly.
+                          Currently deleted down below
                 */
-
-                let target_from = match from.inner {
-                    super::TextureInner::Renderbuffer { raw } => raw,
-                    _ => panic!("Unsupported resolve between non-renderbuffers"),
-                };
-                let target_to = match to.inner {
-                    super::TextureInner::Renderbuffer { raw } => raw,
-                    _ => panic!("Unsupported resolve between non-renderbuffers"),
-                };
-
                 let framebuf_from = gl.create_framebuffer().unwrap();
                 let framebuf_to = gl.create_framebuffer().unwrap();
 
                 gl.bind_framebuffer(glow::READ_FRAMEBUFFER, Some(framebuf_from));
-                gl.framebuffer_renderbuffer(
-                    glow::READ_FRAMEBUFFER,
-                    glow::COLOR_ATTACHMENT0, // NOTE: Assuming color attachment
-                    glow::RENDERBUFFER,
-                    Some(target_from),
-                );
+                match from.inner {
+                    super::TextureInner::Renderbuffer { raw } => {
+                        gl.framebuffer_renderbuffer(
+                            glow::READ_FRAMEBUFFER,
+                            glow::COLOR_ATTACHMENT0, // NOTE: Assuming color attachment
+                            glow::RENDERBUFFER,
+                            Some(raw),
+                        );
+                    }
+                    super::TextureInner::Texture { raw, target } => {
+                        gl.framebuffer_texture_2d(
+                            glow::READ_FRAMEBUFFER,
+                            glow::COLOR_ATTACHMENT0,
+                            target,
+                            Some(raw),
+                            0,
+                        );
+                    }
+                }
 
                 gl.bind_framebuffer(glow::DRAW_FRAMEBUFFER, Some(framebuf_to));
-                gl.framebuffer_renderbuffer(
-                    glow::DRAW_FRAMEBUFFER,
-                    glow::COLOR_ATTACHMENT0, // NOTE: Assuming color attachment
-                    glow::RENDERBUFFER,
-                    Some(target_to),
-                );
-                assert_eq!(
+                match to.inner {
+                    super::TextureInner::Renderbuffer { raw } => {
+                        gl.framebuffer_renderbuffer(
+                            glow::DRAW_FRAMEBUFFER,
+                            glow::COLOR_ATTACHMENT0, // NOTE: Assuming color attachment
+                            glow::RENDERBUFFER,
+                            Some(raw),
+                        );
+                    }
+                    super::TextureInner::Texture { raw, target } => {
+                        gl.framebuffer_texture_2d(
+                            glow::DRAW_FRAMEBUFFER,
+                            glow::COLOR_ATTACHMENT0,
+                            target,
+                            Some(raw),
+                            0,
+                        );
+                    }
+                }
+
+                debug_assert_eq!(
                     gl.check_framebuffer_status(glow::DRAW_FRAMEBUFFER),
                     glow::FRAMEBUFFER_COMPLETE,
                     "DRAW_FRAMEBUFFER is not complete"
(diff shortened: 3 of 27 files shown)


blade-egui/src/lib.rs
```diff
@@ -128,7 +129,11 @@ impl GuiPainter {
     /// It supports renderpasses with only a color attachment,
     /// and this attachment format must be The `output_format`.
     #[profiling::function]
-    pub fn new(info: blade_graphics::SurfaceInfo, context: &blade_graphics::Context) -> Self {
+    pub fn new(
+        info: blade_graphics::SurfaceInfo,
+        context: &blade_graphics::Context,
+        sample_count: u32,
```
kvark, line 135: I don't think this is useful. Egui is used as HUD, and this is practically never rendered in MSAA-enabled surface. MSAA is needed to fight the aliasing of geometry and specularity - neither are a problem for UI in HUD. So the proper thing to do is to always resolve your main renderer, and then draw HUD on top (without multi-sampling).

blade-graphics/src/gles/command.rs
```diff
@@ -806,6 +810,65 @@ impl super::Command {
                     None,
                 );
             }
+
+            Self::BlitFramebuffer { from, to } => {
+                /*
+                    TODO(ErikWDev): Validate
```
kvark, line 816: Please follow these instructions - https://github.com/kvark/blade/tree/main/blade-graphics#opengl-es
and let me know if something doesn't work!

blade-graphics/src/gles/command.rs
```diff
@@ -806,6 +810,65 @@ impl super::Command {
...
+                */
+
+                let target_from = match from.inner {
+                    super::TextureInner::Renderbuffer { raw } => raw,
+                    _ => panic!("Unsupported resolve between non-renderbuffers"),
```
kvark, line 821: Is this just a thing we want to leave for a follow-up, or you are seeing actual technical difficulty here? I think, handling a texture here should be straightforward.

blade-graphics/src/gles/command.rs
```diff
@@ -806,6 +810,65 @@ impl super::Command {
...
+                );
+
+                gl.bind_framebuffer(glow::DRAW_FRAMEBUFFER, Some(framebuf_to));
+                gl.framebuffer_renderbuffer(
+                    glow::DRAW_FRAMEBUFFER,
+                    glow::COLOR_ATTACHMENT0, // NOTE: Assuming color attachment
+                    glow::RENDERBUFFER,
+                    Some(target_to),
+                );
+                assert_eq!(
```
kvark, line 846: let's switch this to `debug_assert_eq!` - no need for the extra GL call in release

blade-graphics/src/gles/command.rs
```diff
@@ -806,6 +810,65 @@ impl super::Command {
...
+                    to.target_size[0] as _,
+                    to.target_size[1] as _,
+                    glow::COLOR_BUFFER_BIT, // NOTE: Assuming color
+                    glow::NEAREST,
+                );
+
+                gl.bind_framebuffer(glow::READ_FRAMEBUFFER, None);
+                gl.bind_framebuffer(glow::DRAW_FRAMEBUFFER, None);
+
+                gl.delete_framebuffer(framebuf_from);
```
kvark, line 868: let's add a TODO: keep the framebuffers owned by the device

blade-graphics/src/gles/resource.rs
```diff
@@ -159,7 +172,11 @@ impl crate::traits::ResourceDevice for super::Context {
                     if desc.array_layer_count > 1 {
                         glow::TEXTURE_2D_ARRAY
```
kvark, line 173: can we support this too? or leaving for a follow-up?
we don't have to do validation in Blade, but it would be polite to at least do a warning
```
log::warn!("Sample count is ignored: not supported for this texture type yet");
```

blade-graphics/src/gles/resource.rs
```diff
@@ -184,13 +201,27 @@ impl crate::traits::ResourceDevice for super::Context {
...
+                            gl.tex_storage_2d(
+                                target,
+                                desc.mip_level_count as i32,
+                                format_desc.internal,
+                                desc.size.width as i32,
+                                desc.size.height as i32,
+                            );
+                        } else {
+                            /*
+                                TODO(ErikWDev): How to set mip count and sample count? Not possible in gles?
```
kvark, line 214: I don't think MSAA makes sense with mipmapping in general. We don't need to support this case - can just assert that the mip count is 1

blade-graphics/src/metal/resource.rs
```diff
@@ -181,7 +181,11 @@ impl crate::traits::ResourceDevice for super::Context {
                 if desc.array_layer_count > 1 {
                     metal::MTLTextureType::D2Array
```
kvark, line 182: same here - do we want to leave this for a followup? if so, let's leave a warning

blade-graphics/src/vulkan/command.rs
```diff
@@ -190,10 +190,34 @@ fn map_render_target(rt: &crate::RenderTarget) -> vk::RenderingAttachmentInfo<'s
...
+            .resolve_mode(vk::ResolveModeFlags::AVERAGE);
+    }
+
+    vk_info.store_op = match rt.finish_op {
+        crate::FinishOp::Store => vk::AttachmentStoreOp::STORE,
+        crate::FinishOp::Discard => vk::AttachmentStoreOp::DONT_CARE,
+        crate::FinishOp::Ignore => vk::AttachmentStoreOp::DONT_CARE,
+        crate::FinishOp::ResolveTo(..) => {
+            /*
+                TODO: DONT_CARE is most optimal in many cases where the msaa texture itself is never read afterwards but only the resolved,
```
kvark, line 211: we can just document `ResolveTo` as leaving the result undefined

blade-graphics/src/vulkan/resource.rs
```diff
@@ -265,13 +265,17 @@ impl crate::traits::ResourceDevice for super::Context {
...
-            samples: vk::SampleCountFlags::from_raw(1), // desc.sample_count
+            samples: vk::SampleCountFlags::from_raw(desc.sample_count),
             tiling: vk::ImageTiling::OPTIMAL,
             usage: map_texture_usage(desc.usage, desc.format.aspects()),
             sharing_mode: vk::SharingMode::EXCLUSIVE,
             ..Default::default()
         };
 
+        /*
+            TODO(ErikWDev): Support lazily allocated texture with transient allocation for efficient msaa?
```
kvark, line 276: I don't think this needs to be brought up in the context of MSAA specifically. Supporting transient allocations would be useful in general.

examples/particle/particle.rs
```diff
@@ -198,7 +202,7 @@ impl System {
         }
     }
 
-    pub fn draw(&self, pass: &mut gpu::RenderCommandEncoder) {
+    pub fn draw(&self, pass: &mut gpu::RenderCommandEncoder, size: (u32, u32)) {
```
kvark, line 205: should we just get the aspect here?

examples/particle/main.rs
```diff
@@ -59,13 +118,44 @@ impl Example {
...
+            array_layer_count: 1,
+            mip_level_count: 1,
+        });
+        let msaa_view = context.create_texture_view(
+            msaa_texture,
+            gpu::TextureViewDesc {
+                name: "msaa texture view",
+                format: surface_info.format,
+                dimension: gpu::ViewDimension::D2,
+                subresources: &gpu::TextureSubresources {
```
kvark, line 141: let's use `&Default::default()`

examples/particle/main.rs
```diff
@@ -4,51 +4,110 @@ use blade_graphics as gpu;
...
+
+        let surface_info = self.surface.info();
+
+        if let Some(sp) = self.prev_sync_point.take() {
+            self.context.wait_for(&sp, !0);
+        }
+        self.context.destroy_texture_view(self.msaa_view);
+        self.context.destroy_texture(self.msaa_texture);
+
+        // self.gui_painter.destroy(&self.context);
```
kvark, line 37: is this commented out because in theory we'd want to recreate the pipelines?
I think, for the purpose of this example, we can just assert that the format hasn't changed. No need to keep the code commented out

blade-graphics/src/metal/resource.rs
```diff
@@ -198,6 +202,7 @@ impl crate::traits::ResourceDevice for super::Context {
             descriptor.set_array_length(desc.array_layer_count as u64);
             descriptor.set_mipmap_level_count(desc.mip_level_count as u64);
             descriptor.set_pixel_format(mtl_format);
+            descriptor.set_sample_count(desc.sample_count as _);
```
kvark, line 205: I don't see where we are handling `ResolveTo` in Metal?

examples/particle/main.rs
```diff
@@ -4,51 +4,110 @@ use blade_graphics as gpu;
 
 mod particle;
 
+// const SAMPLE_COUNT: u32 = 1;
```
kvark, line 7: would be useful to support `SAMPLE_COUNT = 1` in the code here, which would mean - not creating the extra MSAA texture at all

kvark (commented):
This is amazing work, thank you!
High quality, something that any open source project would cherish and greatly appreciate.
Very solid for the first contribution, I'm excited to see this finished.

examples/particle/main.rs
EriKWDev: Usually do it that way. Should I make the SAMPLE_COUNT parameter dynamically changable in the UI as well? Should be possible but the pipeline needs recreating as well in that case

blade-graphics/src/metal/resource.rs
EriKWDev: in `metal/command.rs:223`,

```rust
let store_action = match rt.finish_op {
    crate::FinishOp::Store | crate::FinishOp::Ignore => {
        metal::MTLStoreAction::Store
    }
    crate::FinishOp::Discard => metal::MTLStoreAction::DontCare,
    crate::FinishOp::ResolveTo(ref view) => {
        at_descriptor.set_resolve_texture(Some(view.as_ref()));
        metal::MTLStoreAction::MultisampleResolve
    }
};
at_descriptor.set_store_action(store_action);
```

blade-graphics/src/gles/resource.rs
EriKWDev: Was unsure what to do there, since in that case maybe we should warn about the D3 and D1 case as well, and then about the mip-levels being ignored too. I agree some nice-to-have warnings are appreciated as the user xD Where to stop~

EriKWDev:
> This is amazing work, thank you! High quality, something that any open source project would cherish and greatly appreciate. Very solid for the first contribution, I'm excited to see this finished.

Compared to having created this whole package for us to use and tinker with I don't think its much :) We will continue to evaluate blade for a period and my intention is to contribute if I make good changes.

Can probably fix some of the open winit issues too since have been having to deal with their API changes as well for our game..

examples/particle/particle.rs
EriKWDev: Used the size and sent it directly to the shader before but changed it to be the aspect ratio xP Maybe should be the input to the function now yeah since that is what is used

blade-graphics/src/vulkan/resource.rs
EriKWDev: Yup no not intrinsically related.

It is extra important for bandwidth saving in MSAA on tiled-renderers which is why it popped up in my mind while doing this.

I want to measure though to see what the drivers do with general usage and allocations, though the vulkan samples seems to show with measurements that it makes a bandwidth difference:

https://docs.vulkan.org/samples/latest/samples/performance/msaa/README.html

blade-graphics/src/metal/resource.rs
EriKWDev: Didn't check but yup easily supported case, will add

blade-egui/src/lib.rs
EriKWDev: Please explain a little more. I thought the geometry produced by egui still has edges that could potentially alias. Even if less noticable since the lines are always straight to my knowledge and thus bad cases are mostly avoided, I don't know if rounded corners and stuff are done through geometry or bitmaps.

The docs of eframe mention that one might want to turn on anti-aliasing even if the textures and fonts within egui already are anti-aliased

https://docs.rs/eframe/latest/eframe/struct.NativeOptions.html#structfield.multisampling

EriKWDev:
will re-request review when ready

examples/particle/main.rs
kvark: it's a nice feature but it wouldn't block this PR from landing

blade-egui/src/lib.rs
kvark: I don't think straight horizontal/vertical lines can be aliased. You'd only see the aliasing issue if the stuff is moving, or it's not already aligned to the pixel grid. The advice in their docs is about displaying 3D content within egui, which is arguably a very niche case. Unless your application needs it, let's consider this out of scope and not bother changing blade-egui.

EriKWDev: Yeah while moving, like dragging a window is mostly the case I thought about.

3D embedding would be cool but not something I am doing currently xD

EriKWDev:
For some reason I feel like the egui in blade is much blurrier than I am used to. To the left is our game using wgpu and right is blade

image

blade-egui/src/lib.rs
kvark: If egui is used for the actual game UI, then it's unlikely there is going to be any resizing at all.
If it's used as debug UI. then aliasing of movement doesn't matter, since it's not shipping.
Neither case is also fine with the MSAA overhead for these rectangles.
Smooth edges can be achieved by much simpler means, like analytical anti-aliasing. E.g. we know that this line is barely touching the pixel center so we'll make it more transparent. With this, you get smooth movement without any overhead.

EriKWDev: We are using egui for our entire engine:s editor and have lots of movable windows and whatnot, part of the editor is also intended to be shipped.

Whether masaa for egui is the best and most efficient solution or not is separate :3

blade-graphics/src/gles/command.rs
EriKWDev: Nop, just unfamiliar with opengl. Found framebufferTexture2D so should be easy

blade-graphics/src/gles/command.rs
EriKWDev: Sadly, I am unable to get it to work.

```yaml
[2024-12-07T20:55:05Z ERROR blade_graphics::hal::platform] EGL 'eglCreatePlatformWindowSurface' code 0x300b: _eglCreateWindowSurfaceCommon
thread 'main' panicked at blade-graphics/src/gles/egl.rs:431:26:
called `Result::unwrap()` on an `Err` value: BadNativeWindow
```

I have tried forcing X11 and Wayland and both return the same issue. I am running on  Debian GNU/Linux 12 (bookworm) x86_64 with kernel 6.1.0-28-amd64

blade-graphics/src/gles/command.rs
EriKWDev: Think this should work
```rust
                match from.inner {
                    super::TextureInner::Renderbuffer { raw } => {
                        gl.framebuffer_renderbuffer(
                            glow::READ_FRAMEBUFFER,
                            glow::COLOR_ATTACHMENT0, // NOTE: Assuming color attachment
                            glow::RENDERBUFFER,
                            Some(raw),
                        );
                    }
                    super::TextureInner::Texture { raw, target } => {
                        gl.framebuffer_texture_2d(
                            glow::READ_FRAMEBUFFER,
                            glow::COLOR_ATTACHMENT0,
                            glow::RENDERBUFFER,
                            Some(raw),
                            0,
                        );
                    }
                }
```

blade-graphics/src/gles/command.rs
kvark: Ok, good to know, thank you for trying!
This has probably regressed recently due to #203
I'll check on a few more devices I have...

blade-graphics/src/gles/command.rs
kvark: yep! if in doubt, we can also check wgpu-hal

EriKWDev:
(force push from rebase on main)

Cargo.toml
```diff
@@ -90,9 +90,11 @@ del-geo = "=0.1.29"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 # see https://github.com/emilk/egui/issues/4270
-egui-winit = "0.29"
+egui-winit = { version="0.29", default-features=false, features=["links"] }
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
+# see https://github.com/emilk/egui/issues/4270
+egui-winit = { version="0.29", default-features=false, features=["links"] }
```
kvark, line 97: it's now the same dependency between wasm and normal, can we move it into a common block of dev dependencies?

blade-graphics/src/gles/egl.rs
```diff
@@ -357,7 +357,8 @@ impl super::Context {
                 let window_ptr = unsafe {
                     use objc::{msg_send, runtime::Object, sel, sel_impl};
                     // ns_view always have a layer and don't need to verify that it exists.
-                    let layer: *mut Object = msg_send![handle.ns_view.as_ptr(), layer];
+                    let layer: *mut Object =
+                        msg_send![handle.ns_view.as_ptr() as *mut Object, layer];
```
kvark, line 361: curious, does this cast do anything? I think the type is essentially erased at that point, anyway

kvark (approved):
Looks great, let's get this in!
Just one question though - how do you see this being merged? Do you want to make a pass over commits to have a reasonable (and testable) history, or I can just squash it all together.

blade-graphics/src/gles/egl.rs
EriKWDev: It didn't compile for me, msg_send![] required the thing to implement the "Message" trait and it wasn't implemented by a raw pointer. The cast fixed the compilation though ofc no guarantees

Issue #215, opened 2024-12-06
egui more pixelated than normal?


EriKWDev:
I don't know if it is just me but it looks like egui in blade is more pixelated than I am used to and I don't know what the reason is.

Left is egui with wgpu in our game and right is blade

![image](https://github.com/user-attachments/assets/686b8c25-f6c2-438e-a61e-605251c2b902)

![image](https://github.com/user-attachments/assets/4abdc5b9-ce99-428f-83fd-e059e79d78c5)

Not perfectly scientific

kvark:
The best way to answer this question is to capture these 2 scenarios in something like RenderDoc and share the captures.

One thing that is different is the sampler. Blade is using a single [Linear sampler](https://github.com/kvark/blade/blob/099555282605c7c4cca9e66a8f40148298347f80/blade-egui/src/lib.rs#L161-L167). Wgpu's implementation is more comprehensive and it has a map of different samplers chosen [based on the image](https://github.com/emilk/egui/blob/046034f9020453f1ffe3e96ff26c5404435fcfb5/crates/egui-wgpu/src/renderer.rs#L651-L654).

EriKWDev:
captures, contains the render with current blade main branch, render with blade from #216 as well as render with eframe wgpu

[blade_egui_renderdoc_captures.zip](https://github.com/user-attachments/files/18049777/blade_egui_renderdoc_captures.zip)

kvark:
Thank you for investigation! I left some notes on #216.
It would also help to get a wgpu API trace, so that I can replay this on different platform.

EriKWDev:
> It would also help to get a wgpu API trace, so that I can replay this on different platform.

Trying to create one though egui-wgpu wasn't setup with that feature in mind so had to do some changes there upon wgpu device creation.

There seems to be a wpgu regression at the moment that has removed the trace functionality?
https://github.com/gfx-rs/wgpu/issues/5974

```console
[2024-12-08T09:37:35Z ERROR wgpu::backend::wgpu_core] Feature 'trace' has been removed temporarily, see https://github.com/gfx-rs/wgpu/issues/5974
```

Will see if I can get it to run with older version...

EriKWDev:
I have run into a hoard of different issues with wgpu panicking on destroyed semaphores and egui-winit/egui-wgpu/winit/wgpu versions not being compatible..

Finally have a minimal reproducible that uses a version of wgpu that supports tracing as well as a version of the egui renderer that seems to produce a correct result like eframe.

[...]

PR #216, opened 2024-12-07
Maybe sharper egui soon

EriKWDev:
See notes in #215 

Still not solved but colors are closer. Will keep looking into it

### DIFF
--- a/blade-egui/shader.wgsl
+++ b/blade-egui/shader.wgsl
@@ -79,8 +79,6 @@ fn fs_main_gamma_framebuffer(in: VertexOutput) -> @location(0) vec4<f32> {
 }
 
 
-
-
 // -----------------------------------------------
 // Adapted from
 // https://www.shadertoy.com/view/llVGzG
--- a/blade-egui/shader.wgsl
+++ b/blade-egui/shader.wgsl
@@ -1,3 +1,5 @@
+// NOTE: Borrows heavily from egui-wgpu:s wgsl shaders, used here under the MIT license
+
 struct VertexOutput {
     @location(0) tex_coord: vec2<f32>,
     @location(1) color: vec4<f32>,
@@ -6,8 +8,8 @@ struct VertexOutput {
 
 struct Uniforms {
     screen_size: vec2<f32>,
-    convert_to_linear: f32,
-    padding: f32,
+    dithering: u32,
+    padding: u32,
 };
 var<uniform> r_uniforms: Uniforms;
 
@@ -21,13 +23,6 @@ struct Vertex {
 }
 var<storage, read> r_vertex_data: array<Vertex>;
 
-fn linear_from_srgb(srgb: vec3<f32>) -> vec3<f32> {
-    let cutoff = srgb < vec3<f32>(10.31475);
-    let lower = srgb / vec3<f32>(3294.6);
-    let higher = pow((srgb + vec3<f32>(14.025)) / vec3<f32>(269.025), vec3<f32>(2.4));
-    return select(higher, lower, cutoff);
-}
-
 @vertex
 fn vs_main(
     @builtin(vertex_index) v_index: u32,
@@ -35,8 +30,9 @@ fn vs_main(
     let input = r_vertex_data[v_index];
     var out: VertexOutput;
     out.tex_coord = vec2<f32>(input.tex_coord_x, input.tex_coord_y);
-    let color = unpack4x8unorm(input.color);
-    out.color = vec4<f32>(pow(color.xyz, vec3<f32>(2.2)), color.a);
+    // let color = unpack4x8unorm(input.color);
+    let color = unpack_color(input.color);
+    out.color = color;
     out.position = vec4<f32>(
         2.0 * input.pos_x / r_uniforms.screen_size.x - 1.0,
         1.0 - 2.0 * input.pos_y / r_uniforms.screen_size.y,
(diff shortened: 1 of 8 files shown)


EriKWDev:
investigation progress: https://github.com/kvark/blade/issues/215#issuecomment-2525298153

blade-egui/shader.wgsl
```diff
@@ -50,6 +46,96 @@ var r_texture: texture_2d<f32>;
...
+    if r_uniforms.dithering == 1 {
+        let out_color_gamma_rgb = dither_interleaved(out_color_gamma.rgb, 256.0, in.position);
+        out_color_gamma = vec4<f32>(out_color_gamma_rgb, out_color_gamma.a);
+    }
+    let out_color_linear = linear_from_gamma_rgb(out_color_gamma.rgb);
+    return vec4<f32>(out_color_linear, out_color_gamma.a);
+}
+
+@fragment
+fn fs_main_gamma_framebuffer(in: VertexOutput) -> @location(0) vec4<f32> {
```
kvark, line 67: do we really need a separate entry point? it's largely doing the same thing, we might as well just have an if/else at the end based on a runtime flag

blade-egui/src/lib.rs
```diff
@@ -81,12 +88,62 @@ impl GuiTexture {
...
+                    egui::TextureWrapMode::ClampToEdge => blade_graphics::AddressMode::ClampToEdge,
+                    egui::TextureWrapMode::Repeat => blade_graphics::AddressMode::Repeat,
+                    egui::TextureWrapMode::MirroredRepeat => {
+                        blade_graphics::AddressMode::MirrorRepeat
+                    }
+                };
+                [mode; 3]
+            },
+            mag_filter: match options.magnification {
+                egui::TextureFilter::Nearest => blade_graphics::FilterMode::Nearest,
```
kvark, line 105: we should have helper function to do this mapping, since we are doing it 3 times here

blade-egui/src/lib.rs
```diff
@@ -81,12 +88,62 @@ impl GuiTexture {
...
+                egui::TextureFilter::Linear => blade_graphics::FilterMode::Linear,
+            },
+            min_filter: match options.minification {
+                egui::TextureFilter::Nearest => blade_graphics::FilterMode::Nearest,
+                egui::TextureFilter::Linear => blade_graphics::FilterMode::Linear,
+            },
+            mipmap_filter: match options.mipmap_mode {
```
kvark, line 112: this would look like `options.mipmap_mode.map_or(blade_graphics::FilterMode::Linear, my_mapping_function)`

examples/gui/main.rs
```diff
@@ -0,0 +1,236 @@
+#![allow(irrefutable_let_patterns)]
+
+use blade_graphics as gpu;
+use winit::platform::x11::WindowAttributesExtX11;
+
```
kvark, line 5: do we need a separate example for this? we already have `particle` using egui, `scene` using it, and `move` as well.

blade-egui/shader.wgsl
```diff
@@ -50,6 +46,96 @@ var r_texture: texture_2d<f32>;
...
+    return select(higher, lower, cutoff);
+}
+
+// 0-1 sRGBA gamma  from  0-1 linear
+fn gamma_from_linear_rgba(linear_rgba: vec4<f32>) -> vec4<f32> {
+    return vec4<f32>(gamma_from_linear_rgb(linear_rgba.rgb), linear_rgba.a);
+}
+
+// [u8; 4] SRGB as u32 -> [r, g, b, a] in 0.-1
+fn unpack_color(color: u32) -> vec4<f32> {
```
kvark, line 125: let's use `unpack4x8unorm` builtin instead, no need to reimplement it

blade-egui/src/lib.rs
```diff
@@ -144,7 +204,11 @@ impl GuiPainter {
                 ..Default::default()
             },
             depth_stencil: None, //TODO?
-            fragment: shader.at("fs_main"),
+            fragment: if info.format.is_srgb() {
```
kvark, line 207: This looks highly suspicious to me.
Generally speaking, I expect the rendering pipeline to assume linear space render targets all the way. I don't think there is value in expanding this assumption any further.

EriKWDev: This is what egui-wgpu did so I just replicated all of it to see if it would help.

Egui examples use a non-srg render target whereas blade always uses one

blade-egui/shader.wgsl
EriKWDev: Once I get the same behaviour in the draft I will go back, but yeah this is trivially equivalent.

Wanted to minimize the differences  to egui-wgpu so just took it all over for now

blade-egui/shader.wgsl
EriKWDev: Runtime flag can do it. Just replicated egui-wgpu for the quick draft.

It would be nice to find the underlying cause so the shader potentially can be simple again..

EriKWDev:
Just a note, my approach for the investigation was to minimize differences between blade and egui-wgpu, so I made sure to try and replicate the code.

Once we find the issue we can go back and make only minimal changes :)

examples/gui/main.rs
EriKWDev: Nop, will remove, didnät intend to include it here, was just the one I used to debug

blade-egui/src/lib.rs
kvark: Minimizing changes is a good experiment for sure. Shipping it the way egui_wgpu did seems very concerning to me. I'll try to find some associated discussion there.

EriKWDev:
Closed by #217 and the separate samplers is superseded by #218

PR #217, opened 2024-12-08
egui: switch to gamma space blending, fix blend mode

### LINKED ISSUE #215: egui more pixelated than normal?

kvark:
Fixes #215

Should match https://github.com/emilk/egui/pull/2071 now. With the caveat that upstream is still using an srgb target for some reason, while my PR is switching it to a regular unorm to avoid extra conversion.

Old:
gamma-before

New:
gamma-after2

### DIFF
--- a/blade-egui/shader.wgsl
+++ b/blade-egui/shader.wgsl
@@ -21,10 +21,10 @@ struct Vertex {
 }
 var<storage, read> r_vertex_data: array<Vertex>;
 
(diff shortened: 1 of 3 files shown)


EriKWDev:
Just tried it and it immediately looks more like I am used to, much less pixelated

PR #218, opened 2024-12-08
Egui, one sampler per texture

EriKWDev:
During the egui investigation you mentioned the difference in samplers in blade implementation https://github.com/kvark/blade/issues/215#issuecomment-2524854147.

Extracted the code from the PR #216 that creates a sampler per texture if interesting.

One could create only the sampler permutations requested and lookup based on the options hash during rendering, but this felt simpler

This does not address any issue I have seen, so could also be left as is.


kvark (approved):
Nice, thank you!

PR #224, opened 2024-12-15
Support some more texture formats

### LINKED ISSUE #223: Support more texture formats
I don't know the reasoning behind which formats are currently supported in blade, or if it's just that they've been added on a need-to-have basis

Some formats that we currently use for effects but are missing in blade include
- [x] rg16float
- [x] b10gr11ufloat
- [x] bc6hsfloat + bc6hufloat

Some formats we would like to experiment with
- [x] rgb9e5float
- [x] rgb10a2unorm / bgr10a2unorm
- [x] bc7

Furthermore intressed in exploring, maybe out of scope now
- [ ] astc
- [ ] etc2 

All these (bar astc/etc2) are fairly supported according to gpuinfo. They all exist in vulkan and metal, haven't checked opengl-es.

I suspect more formats could be supported but we don't require more atm

EriKWDev:
Closes #223 

Wasn't quite sure with the naming scheme for the formats so if you have a better capitalization or declaration order feel free to change or come with opinion and I'll change

Translated the gles-texture-descriptions from wgpu-hal though I don't know how to describe the BGR10_A2 format so just did RGBA10_A2 there as well and left a todo. 

I also noticed that some 16f formats in gles seemed incorrectly labeled 'FLOAT' when they according to wgpu-hal should have been 'HALF_FLOAT'

### DIFF
--- a/blade-graphics/src/lib.rs
+++ b/blade-graphics/src/lib.rs
@@ -309,7 +309,7 @@ pub enum TextureFormat {
     Bc5Unorm,
     Bc5Snorm,
     Bc6hUfloat,
-    Bc6hSfloat,
+    Bc6hFloat,
     Bc7Unorm,
     Bc7UnormSrgb,
     // packed 32-bit
--- a/blade-graphics/src/lib.rs
+++ b/blade-graphics/src/lib.rs
@@ -308,15 +308,15 @@ pub enum TextureFormat {
     Bc4Snorm,
     Bc5Unorm,
     Bc5Snorm,
-    Bc6UFloat,
-    Bc6SFloat,
+    Bc6hUfloat,
+    Bc6hSfloat,
     Bc7Unorm,
     Bc7UnormSrgb,
     // packed 32-bit
     Rgb10a2Unorm,
     Bgr10a2Unorm,
     Rg11b10Float,
-    Rgb9e5Float,
+    Rgb9e5Ufloat,
 }
 
 #[derive(Clone, Copy, Debug)]
--- a/blade-graphics/src/lib.rs
+++ b/blade-graphics/src/lib.rs
@@ -287,6 +287,7 @@ pub enum TextureFormat {
     Bgra8UnormSrgb,
     Rgba8Snorm,
     R16Float,
+    Rg16Float,
     Rgba16Float,
     R32Float,
     Rg32Float,
@@ -307,6 +308,15 @@ pub enum TextureFormat {
     Bc4Snorm,
     Bc5Unorm,
     Bc5Snorm,
+    Bc6UFloat,
+    Bc6SFloat,
+    Bc7Unorm,
+    Bc7UnormSrgb,
+    // packed 32-bit
+    Rgb10a2Unorm,
+    Bgr10a2Unorm,
+    Rg11b10Float,
+    Rgb9e5Float,
 }
 
 #[derive(Clone, Copy, Debug)]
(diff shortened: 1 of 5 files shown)


blade-graphics/src/lib.rs
```diff
@@ -307,6 +308,15 @@ pub enum TextureFormat {
     Bc4Snorm,
     Bc5Unorm,
     Bc5Snorm,
+    Bc6hUfloat,
+    Bc6hFloat,
+    Bc7Unorm,
+    Bc7UnormSrgb,
+    // packed 32-bit
+    Rgb10a2Unorm,
+    Bgr10a2Unorm,
```
kvark, line 317: let's not add this one. It's not very useful, and it's not translating nicely

kvark (commented):
Thank you!
Just one small correction, otherwise looks great

blade-graphics/src/vulkan/mod.rs
```diff
@@ -556,6 +557,14 @@ fn map_texture_format(format: crate::TextureFormat) -> vk::Format {
         Tf::Bc4Snorm => vk::Format::BC4_SNORM_BLOCK,
         Tf::Bc5Unorm => vk::Format::BC5_UNORM_BLOCK,
         Tf::Bc5Snorm => vk::Format::BC5_SNORM_BLOCK,
+        Tf::Bc6hUfloat => vk::Format::BC6H_UFLOAT_BLOCK,
+        Tf::Bc6hFloat => vk::Format::BC6H_SFLOAT_BLOCK,
+        Tf::Bc7Unorm => vk::Format::BC7_UNORM_BLOCK,
+        Tf::Bc7UnormSrgb => vk::Format::BC7_SRGB_BLOCK,
+        Tf::Rgb10a2Unorm => vk::Format::A2R10G10B10_UNORM_PACK32,
```
kvark, line 564: also, this needs to be `A2B10G10R10_UNORM_PACK32`

EriKWDev: So a texture called Rgb should be Bgr? I'm confused.

I saw that BGR was more supported for this format on gpuinfo xD

Do you mean that the distinction is not important to the user of blade so it's better to have one and choose the most adopted?

kvark: In vulkan, the formats that have "_PACK??" at the end are Big Endian encoded. So for all practical purposes, if you see PACK32, treat the components in the reverse order. Thus, our `Rgb10a2Unorm` would be equivalent to `A2G10B10A10_UNORM_PACK32` in Vulkan.

EriKWDev: Oh, how intuitive! 

Well that explains why the bgra format on gpuinfo was more supported 

Thanks for the explanation and will fix

blade-graphics/src/lib.rs
EriKWDev: removing

blade-graphics/src/vulkan/mod.rs
EriKWDev: As for metal, it is correct as is?

Seems wgpu-hal in `metal/adapter.rs` does "Tf::Rgb10a2Uint => RGB10A2Uint,"

kvark: Yes, Metal is a sane API :)

kvark:
Do you mind if I squash this? Or you want to squash yourself?

kvark (approved):

EriKWDev:
Do it :)

PR #225, opened 2024-12-15
Fix particle example

EriKWDev:
The particle system was not re-initialized upon recreation. This caused black screen rendering on macOS after change as well as undefined buffer content on vulkan after reset


kvark (approved):
thank you!

PR #228, opened 2024-12-19
Optional fragment shader

### LINKED ISSUE #226: Ability to describe render-pipelines without a fragment shader
The `RenderPipelineDesc` has a mandatory `pub fragment: ShaderFunction<'a>` field.

For a vertex-only shader it doesn't make sense to require the declaration of a fragment shader

I propose it change to a `pub fragment: Option<ShaderFunction<'a>>` instead

EriKWDev:
Closes #226 

This changes the declaration of the fragment function from being mandatory to instead be optional.

Vulkan and Metal have been tested in the sense that bunnymark with fragment shader set to None compiles and runs. Upon inspection of vulkan run of the bunnymark example without fragment specified, I observed that the vertex shader is executed correctly but renderdoc states there is no fragment shader stage as expected.

### DIFF
--- a/blade-egui/src/lib.rs
+++ b/blade-egui/src/lib.rs
@@ -183,7 +183,7 @@ impl GuiPainter {
                 ..Default::default()
             },
             depth_stencil: None, //TODO?
-            fragment: shader.at("fs_main"),
+            fragment: Some(shader.at("fs_main")),
             color_targets: &[blade_graphics::ColorTargetState {
                 format: info.format,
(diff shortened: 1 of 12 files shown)


EriKWDev:
I can still not get the gles backend to run so I just assume in this PR it is fine to not create the fragment shader in the pipeline.

Don't know if it is required to supply an empty fragment shader

kvark (approved):
Thank you!

PR #230, opened 2024-12-23
Merge instances of set viewport/scissor at the API level

kvark:
Following up to #227

Reducing code and API duplication

### DIFF
--- a/blade-graphics/src/vulkan/mod.rs
+++ b/blade-graphics/src/vulkan/mod.rs
@@ -252,6 +252,7 @@ impl ComputePipeline {
     }
 }
 
+#[derive(Debug)]
 pub struct RenderPipeline {
     raw: vk::Pipeline,
     layout: PipelineLayout,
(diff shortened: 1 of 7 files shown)


blade-graphics/src/vulkan/mod.rs
```diff
@@ -295,11 +296,15 @@ pub struct ComputeCommandEncoder<'a> {
     device: &'a Device,
     update_data: &'a mut Vec<u8>,
 }
+//Note: we aren't merging this with `ComputeCommandEncoder`
+// because the destructors are different, and they can't be specialized
+// https://github.com/rust-lang/rust/issues/46893
```
EriKWDev, line 301: Didn't know about this one

kvark: Quite annoying, honestly! It breaks the API symmetry between standalone types and generic specializations.
I've spent some time experimenting with it and had to throw a lot of stuff out, due to this seemingly minor issue.

EriKWDev: Very surprising it isn't supported :/

PR #233, opened 2024-12-25
make traits mod public

EriKWDev:
While keeping the traits opaque to the user using the current `hidden_trait` proc-macro makes it very simple for the user, having access to the traits in order to write utility-functions that compile on all backends is still a nice ability.

I read the `hidden_trait` code and I believe making this public would still work, get a concrete function implementation as well as the trait implementation available if needed.

Arose from: https://github.com/kvark/blade/pull/227#issuecomment-2560604563


EriKWDev:
🎄🎅

kvark (approved):

PR #239, opened 2024-12-27
vulkan: shift viewport y by height

### LINKED ISSUE #238: flip_y_requires_shift
I have just spent a some time debugging why my `set_viewport` calls seem to cause inconsistent behaviour between metal on macos and my vulkan linux. After some investigation into `wgpu-hal` and `blade`, I found a difference in how viewports are created. 

In blade, the viewport in vulkan always has the `y` set to the height upon starting a render
```rust
/// blade:s begin_render
let viewport = vk::Viewport {
    x: 0.0,
    y: target_size[1] as f32, // AHA! Missed this.. We shift it here already
    width: target_size[0] as f32,
    height: -(target_size[1] as f32),
    min_depth: 0.0,
    max_depth: 1.0,
};
```
but wgpu-hal sometimes has it set to 0
```rust
/// wgpu-hal:s, begin_render
let vk_viewports = [vk::Viewport {
    x: 0.0,
    y: if self.device.private_caps.flip_y_requires_shift {
        desc.extent.height as f32
    } else {
        0.0
    },
    width: desc.extent.width as f32,
    height: -(desc.extent.height as f32),
    min_depth: 0.0,
    max_depth: 1.0,
}];
```

```rust
let vk_viewports = [vk::Viewport {
    x: rect.x,
    y: if self.device.private_caps.flip_y_requires_shift {
        rect.y + rect.h // shift is applied..
    } else {
        rect.y
    },
    width: rect.w,
    height: -rect.h, // flip Y
    min_depth: depth_range.start,
    max_depth: depth_range.end,
}];
```

In wgpu-hal `vulkan/mod.rs` they state:
```rust
struct PrivateCapabilities {
    /// Y-flipping is implemented with either `VK_AMD_negative_viewport_height` or `VK_KHR_maintenance1`/1.1+. The AMD extension for negative viewport height does not require a Y shift.
    ///
    /// This flag is `true` if the device has `VK_KHR_maintenance1`/1.1+ and `false` otherwise (i.e. in the case of `VK_AMD_negative_viewport_height`).
    flip_y_requires_shift: bool,
    // ..
}
```

EriKWDev:
Closes #238 (I think)

Since blade requires vulkan >= 1.1, I understand it as this is should always be done?

### DIFF
--- a/blade-graphics/src/vulkan/command.rs
+++ b/blade-graphics/src/vulkan/command.rs
@@ -857,7 +857,7 @@ impl crate::Viewport {
     fn to_vk(&self) -> vk::Viewport {
         vk::Viewport {
             x: self.x,
-            y: self.y,
+            y: self.y + self.h, // shift since always flipped y and vulkan >= 1.1
             width: self.w,
             height: -self.h, // flip Y
             min_depth: self.depth.start,
--- a/blade-graphics/src/vulkan/command.rs
+++ b/blade-graphics/src/vulkan/command.rs
@@ -389,21 +389,21 @@ impl super::CommandEncoder {
             }
         }
 
-        let render_area = vk::Rect2D {
-            offset: Default::default(),
-            extent: vk::Extent2D {
-                width: target_size[0] as u32,
-                height: target_size[1] as u32,
-            },
-        };
-        let viewport = vk::Viewport {
+        let render_area = crate::ScissorRect {
+            x: 0,
+            y: 0,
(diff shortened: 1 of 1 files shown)


EriKWDev:
<details>
  <summary>renderer port from wgpu to blade progress</summary>

image

stable cascaded shadow maps and compute-shader frustum-culling with indirect draw now done!

</details>

kvark (approved):
Oh snap, how did we miss that!

PR #240, opened 2024-12-28
vulkan: missing to account for cull_mode

EriKWDev:
`cull_mode` was not being forwarded in vulkan impl


EriKWDev:
don't think the gles impl does it either?

kvark (approved):
Nice, thank you!

PR #242, opened 2025-01-05
Stencil Additions

EriKWDev:
### Motivation
We want to use stencil tests to optimize part of our planar reflection rendering

### Added Texture Formats
- Depth32FloatStencil8Uint
- Stencil8Uint

### API Additions
Added `set_stencil_reference(&mut self, reference: u32)` to `RenderEncoder` trait
Both metal, gles and vulkan support setting it for front and back faces individually, but from what I could find metal only supports setting them both at the same time.

I decided to simply set front and back to same value for now. Could be solved by requiring both be specified as a `front_reference: u32` and `back_reference: u32`. Skipped for now

### Metal
Noticed that metal wasn't doing anything with the stencil load/clear ops so implemented those while peeking at `wpgu-hal` implementation 

### Testing
Tested and inspected in vulkan. Gles compiles on linux. Macos compiles and runs.


EriKWDev:
The results using stencil test for planar reflection show very good results. We see performance gain on our weaker laptop targets.

EriKWDev:
One thing that caught me was that the `TextureColor` became a little unintuitive when I wanted to clear the depth and stencils to 1.0 and 0.

`OpaqueBlack` needed a few extra brain cells for me to associate xD But makes sense once you rule out the other options of what it could mean

blade-graphics/src/util.rs
```diff
@@ -74,6 +74,8 @@ impl super::TextureFormat {
             Self::Rg32Uint => uncompressed(8),
             Self::Rgba32Uint => uncompressed(16),
             Self::Depth32Float => uncompressed(4),
+            Self::Depth32FloatStencil8Uint => uncompressed(5),
```
kvark, line 77: both depth and stencil are not necessarily stored together, and unlikely really occupy 5 bytes (more like 8 in this case)
let's leave a comment here that this value isn't really applicable. We should probably just set it to 0, even, to avoid confusion

kvark (approved):
Nice work!

blade-graphics/src/util.rs
EriKWDev: will add comment in next PR

PR #243, opened 2025-01-09
Add `supports_texture_sample_count` method to Metal and Vulkan.

huacnlee:
There need a method to check if the system supports a specific sample count.

Ref https://github.com/zed-industries/zed/pull/22812 need this feature to setup sample_count.


huacnlee:
```bash
cargo run --example particle
```

Tested `particle` example on Windows and macOS, is looks worked will.

EriKWDev:
Nice addition!

The egui dropdown for selecting the MSAA count could perhaps use the list of supported formats? Currently hard-coded to:
```rust
for i in [1, 2, 4] {
// ...
}
```

huacnlee:
You're right, I changed it.

kvark (approved):
Thank you!

PR #244, opened 2025-01-10
Add and implement dispatch indirect

EriKWDev:
Existed in Vulkan since 1.0, in metal and in opengl ES since 3.1.

Gles compiles on linux, tested on macos and linux with vulkan.

Motivation is that we are doing compute-based culling and would like to dispatch a sorting compute afterwards. The input to the sort function depends on the results from the cull. `dispatch_indirect` allows us to avoid dispatching unnecessarily many groups or doing a cpu read-back


kvark (approved):
you are unstoppable, thank you!

Issue #246, opened 2025-01-12
Maybe find other solution than TextureColor for specifying clear values


EriKWDev:
I see there seem to be *three* mixups in the current implementation as to what depth and stencil values the TextureColors should map to xD

the *metal* version does this
```console
name                depth  stencil
TransparentBlack    0.0       0
OpaqueBlack         0.0       0 [*]
White               1.0      !0
```

the *vulkan* version does this
```console
name                depth  stencil
TransparentBlack    0.0       0
OpaqueBlack         1.0 [*]   0 [*]
White               1.0      !0
```

I think the logical mapping is that transparent/opaque maps to the stencil 0/!0 and black/white maps to 0.0/1.0

If that reasoning is correct, and given that 'white' implies opaque, the canonical table should be
```console
name                depth  stencil
TransparentBlack    0.0       0
OpaqueBlack         0.0      !0
White               1.0      !0
```

There are some things that cause me to think we should reconsider this:
  - usually I think of 'depth then stencil' in that order, now transparency (first word) maps to stencil and color to the depth, so a flip has to happen
  - cannot represent depth=1.0, stencil=0 ("TransparentWhite"). For a projection without reverse-z, this is a useful case
  - cannot utilize full stencil as only all bits set to 1 or 0, no masks possible
  - cannot represent arbitrary clear color for color targets

I can see there is a cleanliness argument for being able to represent most things with these three, but I think it is worth a reconsideration, especially for depth and stencil, but also for color to allow for arbitrary clear color which is supported by all backends

kvark:
Yeah, we need a way to provide exact depth and stencil clear values. The current enum is only useful for the color...

EriKWDev:
The current PR #247 just fixes the implementations so they at least do the same, but I would indeed prefer a change in how it is expressed :)

Do you have a suggestion? wgpu has the StoreOp be generic over a value. 

To my understanding it is not possible to have separate load and finish options for depth/stencil in vulkan with dynamic rendering (why?)? I saw it is possible in Metal and with the normal `vk::AttachmentDescription` but not in `vk::RenderingAttachmentInfo`

It would have been useful to us actually, but alas. Did a full screen triangle overwriting depth with stencil set to read only so managed to do it xP

kvark:
suggestion
```
pub enum InitOp<V> {
    Load,
    Clear(V),
    DontCare,
}

pub struct DepthStencilValue {
  depth: f32,
  stencil: u8,
}

[...]

PR #247, opened 2025-01-12
stencil/depth unification

EriKWDev:
per #246 the depths and stencils were not matching up between the backends, and was inconsistent with naming. I moved the declaration of these clear values to a single spot to avoid this.

I think I would've preferred to just specify them manually though


EriKWDev:
I am using this one in the meantime to have standardized behavior at least xD Let's decide on a way to make the values specifiable in the API

kvark (approved):

PR #252, opened 2025-02-11
Adds support for external buffers and images

ewoudje:
The use I'm using it for is to have external images using a Win32 handle. Because this is very platform specific and niche I instead made a way to import self-made images and memory.

Specifically i exposed the vulkan device, the `Allocation` struct, `allocate_memory` and `free_memory`.
I added 2 functions to the vulkan context for importing memory and textures.
And added `select_memory_type` as a utility function.

I also added a way to add extensions when creating the context using a new struct `PlatformSpecificContextDesc`.
Which is used by the new function `init_with_platform_desc`. The original init function is still the same so it doesn't break anything.

### DIFF
--- a/blade-graphics/src/vulkan/resource.rs
+++ b/blade-graphics/src/vulkan/resource.rs
@@ -7,6 +7,7 @@ struct Allocation {
     offset: u64,
     data: *mut u8,
     handle: usize,
+    memory_type: crate::Memory
 }
 
 impl super::Context {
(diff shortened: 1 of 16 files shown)


kvark:
Nice! I was working on the exact same thing today in another project and thinking that we need this in Blade.
Let me have a good look at it.

kvark:
I would approach this differently. Instead of exposing the memory infrastructure and device creation facilities, we could allow buffer and texture creation to support external file descriptors. For buffers specifically, the API is very nice:
```rust
#[derive(Clone, Copy, Debug, PartialEq)]
pub enum Memory {
    Device,
    Shared,
    Upload,
    External { import_fd: i32 }, // our new variant
}
```
More generally, we'd even go with `import_fd: Option<if32>` such as `None` corresponds to memory that is exported from Blade but not imported from the outside.

For textures, it's a bit less clean since there is no enum like `Memory` used there. We could do something like this:
```rust
struct ExternalMemory {
  import_fd: Option<i32>,
}
struct TextureDesc {
  ...
  external: Option<ExternalMemory>,
}
```

Note that all the sharing is done via a file descriptor in this case. Besides buffers and textures, there is one more thing that is possibly complicated: synchronization. We need a way to import and export a semaphore.

ewoudje:
> I would approach this differently. Instead of exposing the memory infrastructure and device creation facilities, we could allow buffer and texture creation to support external file descriptors. For buffers specifically, the API is very nice:
> 
> ```rust
> #[derive(Clone, Copy, Debug, PartialEq)]
> pub enum Memory {
>     Device,
>     Shared,
>     Upload,
>     External { import_fd: i32 }, // our new variant
> }
> ```
> 
> More generally, we'd even go with `import_fd: Option<if32>` such as `None` corresponds to memory that is exported from Blade but not imported from the outside.
> 
> For textures, it's a bit less clean since there is no enum like `Memory` used there. We could do something like this:
> 
> ```rust
> struct ExternalMemory {
>   import_fd: Option<i32>,
> }
> struct TextureDesc {
>   ...
>   external: Option<ExternalMemory>,
> }
> ```
> 
> Note that all the sharing is done via a file descriptor in this case. Besides buffers and textures, there is one more thing that is possibly complicated: synchronization. We need a way to import and export a semaphore.

How about having in struct `TextureDesc` a field for the enum `Memory`?
It can then simply be fed into `let allocation = self.allocate_memory(requirements, crate::Memory::Device);` replacing the hardcoded device memory part.

ewoudje:
It seems github closed it because discarded previous commits

ewoudje:
Should now be complete, i don't have a eqv metal implementation since ive never used metal and also just don't have a machine that runs metal. Same for the metal external sources.

I also haven't added a eqv gles impl. But it seems that it currently is not working? Or at least i couldn't get it to work without modifying it quite a bit.

I added a boolean in the mirror example that can be set true to test exporting a image, But you would need another application to import it.

kvark:
Sorry, I haven't been tracking this. Will get you some feedback over the weekend. Thank you for pushing this!

blade-graphics/src/vulkan/resource.rs
```diff
@@ -83,6 +156,35 @@ impl super::Context {
...
+        let block = manager
+            .slab
+            .get(texture.memory_handle)
+            .expect("get_texture_fd: Invalid memory_handle");
+
+        //TODO make non-windows variant
+        #[cfg(target_os = "windows")]
+        {
+            let get_handle_info = vk::MemoryGetWin32HandleInfoKHR {
+                handle_type: vk::ExternalMemoryHandleTypeFlags::OPAQUE_WIN32_KMT,
```
kvark, line 170: any reason this is hardcoded to KMT?

blade-graphics/src/vulkan/resource.rs
```diff
@@ -192,6 +294,14 @@ impl crate::traits::ResourceDevice for super::Context {
 
     fn create_buffer(&self, desc: crate::BufferDesc) -> super::Buffer {
         use vk::BufferUsageFlags as Buf;
+        let external_next = match &desc.memory {
```
kvark, line 297: Please don't match on a reference. Nothing wrong with it in the community at large, but I don't want this pattern used in my projects.

blade-graphics/src/vulkan/resource.rs
```diff
@@ -208,7 +318,17 @@ impl crate::traits::ResourceDevice for super::Context {
                 Buf::SHADER_DEVICE_ADDRESS | Buf::ACCELERATION_STRUCTURE_BUILD_INPUT_READ_ONLY_KHR;
         }
 
-        let raw = unsafe { self.device.core.create_buffer(&vk_info, None).unwrap() };
+        let raw = unsafe {
+            if let Some(mut e) = external_next {
+                self.device
+                    .core
+                    .create_buffer(&vk_info.push_next(&mut e), None)
```
kvark, line 325: let's only do the `push_next` thing inside the `if`, and have `create_buffer` call one and only

blade-graphics/src/vulkan/resource.rs
```diff
@@ -276,9 +401,23 @@ impl crate::traits::ResourceDevice for super::Context {
...
                             Measure bandwidth usage!
         */
-        let raw = unsafe { self.device.core.create_image(&vk_info, None).unwrap() };
+        let raw = unsafe {
+            if let Some(mut e) = external_next {
+                self.device
+                    .core
+                    .create_image(&vk_info.push_next(&mut e), None)
+            } else {
+                self.device.core.create_image(&vk_info, None)
```
kvark, line 410: same here

blade-graphics/src/vulkan/resource.rs
```diff
@@ -276,9 +401,23 @@ impl crate::traits::ResourceDevice for super::Context {
...
+            }
+            .unwrap()
+        };
+
         let requirements = unsafe { self.device.core.get_image_memory_requirements(raw) };
-        let allocation = self.allocate_memory(requirements, crate::Memory::Device);
+        let allocation = self.allocate_memory(
+            requirements,
+            desc.external
+                .map_or(crate::Memory::Device, |e| crate::Memory::External(e)),
```
kvark, line 419: nit: `map_or(crate::Memory::Device, crate::Memory::External)`

examples/bunnymark/main.rs
```diff
@@ -1,6 +1,7 @@
 #![allow(irrefutable_let_patterns)]
 
 use blade_graphics as gpu;
+use blade_graphics::Memory;
```
kvark, line 4: please follow the existing style everywhere!
all the `blade_graphics` stuff is accessed as `gpu::` here

examples/particle/main.rs
```diff
@@ -183,7 +201,7 @@ impl Example {
             .update_textures(&mut self.command_encoder, gui_textures, &self.context);
         self.particle_system.update(&mut self.command_encoder);
 
-        if self.sample_count <= 1 {
+        if self.sample_count <= 1 && !self.export_image {
```
kvark, line 204: I don't understand this part. Are we going the MSAA route whenever `self.export_image` is true?

blade-render/src/render/mod.rs
```diff
@@ -8,6 +8,7 @@ pub use debug::{DebugBlit, DebugLine, DebugPoint};
 pub use dummy::DummyResources;
 pub use env_map::EnvironmentMap;
 
+use blade_graphics::Memory;
```
kvark, line 11: similarly, please follow the existing style in the modules

blade-render/src/render/env_map.rs
```diff
@@ -1,6 +1,6 @@
-use std::num::NonZeroU32;
-
 use crate::DummyResources;
+use blade_graphics::Memory;
```
kvark, line 2: same here

blade-render/src/render/dummy.rs
```diff
@@ -1,3 +1,4 @@
+use blade_graphics::Memory;
```
kvark, line 1: and here

blade-graphics/src/vulkan/resource.rs
```diff
@@ -1,5 +1,8 @@
+use crate::hal::Device;
+use crate::ExternalMemorySource;
```
kvark, line 2: do we really need these new use statements?

blade-graphics/src/vulkan/resource.rs
```diff
@@ -1,5 +1,8 @@
...
 use ash::vk;
 use gpu_alloc_ash::AshMemoryDevice;
+use std::marker::PhantomData;
```
kvark, line 5: please follow the existing style - there is one `use` statement per crate.

examples/bunnymark/main.rs
ewoudje: Yhea my bad, i tried doing that, but it seems i accidentally had my editor import still some. I should probably turn that off.😅

examples/particle/main.rs
ewoudje: Yes since the MSAA image is configured to export (if enabled). So it would have to render to that for it to be exported even if its only sampled once.

blade-graphics/src/vulkan/resource.rs
ewoudje: Ah, i added this quickly so i could test in the mirror example, but i should indeed provide a way for each external type.

ewoudje: If its alright for the return value I'm gonna use `ExternalMemorySource` although its a bit off that it returns it with always `Some` in the enums. It would at least make it so there doesn't have to be another enum full off platform specific configurations.

blade-graphics/src/vulkan/resource.rs
kvark: It relies on more sugar in the syntax without clear benefits. When looking at the match statement, I want to know exactly what's referenced and what's copied. Without the sugar, I can look at the statement, look at the type definition, and get that answer. With sugar, I need to consider more things.

examples/bunnymark/main.rs
kvark: Still there?

kvark (commented):
Thank you for patience! I think we can land after the last pass - could you determine the commit structure you want after landing? I.e. do you want to squash some things, or prefer me to squash the whole PR?

ewoudje:
Alright removed the import that i forgot, and fixed another CI error.
And yeah, squash the whole PR, since i use multiple machines i made a bunch of in-between commits that aren't very relevant/informative.

kvark:
Looks like there is a build failure on macos

Issue #255, opened 2025-04-19
Is VK_KHR_external_memory[_fd, _win32] required?


hmartinez82:
I'm trying to run blade with Mesa's Lavapipe, but on Windows it implements only `VK_KHR_external_memory`, but not `VK_KHR_external_memory_win32``.

Would it be possible to add them to the list of required Vulkan extensions to https://crates.io/crates/blade-graphics ?

Lavapipe seems to implement all the other extensions in the Readme.

Are there other extensions not listed there?

Thank you 🙏

kvark:
Oh that's interesting. We assume that the platform-specific side is provided given the base extension support.
It's strange that lavapipe doesn't expose it like this.
@ewoudje any recommendations?

hmartinez82:
It does not :( https://gitlab.freedesktop.org/mesa/mesa/-/issues/13017

ewoudje:
So if i understand correctly, the problem is that blade detects that `VK_KHR_external_memory` is supported, and then assumes  `VK_KHR_external_memory_win32` is supported because we are on windows.

So the easy fix would be just also checking the platform specific extensions, this does mean that you wouldn't be able to use blade with external memory support on lavapipe on windows. Which would fix the OPs error as zed doesn't use external memory. 

The not so easy fix is throwing out the whole assumption that fd is for unix based OSes and win32 is for windows and only look at the extensions themselves.  As a cons this would remove the compiletime check for using windows enums on linux targets etc.

The middleground would be the easy fix but then adding a `or` in the cfg attributes to allow endusers to specify a flag to override OS assumptions. Which would be the best of both worlds.

hmartinez82:
Is that doable? That would be awesome. Sucks I can't help because I'm not a rust developer nor a vulkan dev either 😅

kvark:
I think we want is "just also checking the platform specific extensions".
@ewoudje would you be kind enough to make this patch?

andrejohansson:
I see the PR, does this mean that this issue is resolved? and does it help with #222?

hmartinez82:
@andrejohansson I hope so, that's the reason behind this request of mine :)

ewoudje:
> [@ewoudje](https://github.com/ewoudje) I built blade with your changes, but now there's something new. 😔, that once again doesn't happen in Linux. Now, instead of complaining about a missing extension, when I run: `cargo run --release --example bunnymark` with Lavapipe. I get:
> 
> ```
[...]

PR #258, opened 2025-05-14
Fix external_memory access when not needed

ewoudje:
Moved getting the external_memory device into the match itself.
This should be the only spot where i access this device, so the same kind of oopsie shouldn't be at another spot.
Fixes https://github.com/kvark/blade/issues/255#issuecomment-2877408678


kvark (approved):

kvark:
nice!

PR #263, opened 2025-05-26
fix: draw_indirect & draw_indexed_indirect

sunli829:
The `draw_indirect` and `draw_indexed_indirect` functions were missing the `draw_count` parameter. I added it, and now these two functions are working correctly.

Additionally, I copied the `DrawIndirectArgs` and `DrawIndexedIndirectArgs` structs from the `wgpu` project.

### DIFF
--- a/blade-graphics/src/gles/command.rs
+++ b/blade-graphics/src/gles/command.rs
@@ -537,7 +537,7 @@ impl crate::traits::RenderPipelineEncoder for super::PipelineEncoder<'_> {
         });
     }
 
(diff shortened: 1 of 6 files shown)


kvark (commented):
The design choice to not expose the count was a bit intentional. Only Vulkan supports it, and it's always trivial to do it on your side with a loop. So it doesn't expose any superpowers that are outside of easy reach.

sunli829:
Thank you for telling me, using loops indeed made no difference in performance.

PR #266, opened 2025-07-11
Return error on out of date surface

cetra3:
I've ran into an issue using Zed in a Tiling WM that it gets stuck in a loop and never resolves itself.  This is because there is something invalidating the surface, but not sure what it is exactly.  The fix is to, when this happens, reinitialise the surface.

This PR makes the `acquire_frame` return a `Result` so the callee can handle this case correctly:  VK_ERROR_OUT_OF_DATE_KHR: The swap chain has become incompatible with the surface and can no longer be used for rendering. Usually happens after a window resize, but there seems to be other reasons for it.

### DIFF
--- a/blade-graphics/src/vulkan/surface.rs
+++ b/blade-graphics/src/vulkan/surface.rs
@@ -19,26 +19,8 @@ impl super::Surface {
     }
 
     pub fn acquire_frame(&mut self) -> super::Frame {
-        let acquire_semaphore = self.next_semaphore;
-        match unsafe {
-            self.device.acquire_next_image(
-                self.swapchain.raw,
-                !0,
-                acquire_semaphore,
-                vk::Fence::null(),
-            )
-        } {
-            Ok((index, _suboptimal)) => {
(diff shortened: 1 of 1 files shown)


cetra3:
How about a new method that returns a `Result` and this method calls that method internally?

That way the signature is kept the same but we can still get the error out if we want to

cetra3:
I've adjusted the PR to include a `try_acquire_frame` method, but keeping the semantics/signature of `acquire_frame` intact

kvark:
> how do I check why a surface is out of date without this change

With current code, the only error that is not fatal is OUT_OF_DATE, in which case we'd return a "dummy" frame. Admittedly, the user can't know if it's dummy. We could expose a simple method, e.g. `is_valid()`, for the user to know.

The approach you are suggesting - via `try_acquire_frame` is fine. The only problem is that it's not platform-agnostic.
We'd either need to define an error type that is platform-agnostic and implement this for other backends, or we add "vulkan" somewhere in the function name to make it clear that this is backend-specific.

PR #291, opened 2025-12-19
Add support for dual-source blending

reflectronic:


blade-graphics/src/vulkan/pipeline.rs
```diff
@@ -568,6 +568,12 @@ impl crate::traits::ShaderDevice for super::Context {
             let mut vk_attachment = vk::PipelineColorBlendAttachmentState::default()
                 .color_write_mask(vk::ColorComponentFlags::from_raw(ct.write_mask.bits()));
             if let Some(ref blend) = ct.blend {
+                let uses_dual_source = blend_component_uses_dual_source(&blend.color)
+                    || blend_component_uses_dual_source(&blend.alpha);
+                if uses_dual_source && !self.dual_source_blending {
```
kvark: nit: it's essentially an `assert!`

kvark (approved):

blade-graphics/src/metal/mod.rs
```diff
@@ -532,6 +532,7 @@ impl Context {
                 .map(|v| 1 << v)
                 .filter(|&count| device.supportsTextureSampleCount(count as _))
                 .sum(),
+            dual_source_blending: true,
```
reflectronic: It seems like dual-source blending is supported on Apple1 and Mac1 families, so, I believe `true` is fine here.

reflectronic:
Thanks!!

PR #339, opened 2026-04-14
Multi-Queue Support

### LINKED ISSUE #329: Support for multiple submission queues? (async compute)
`blade-graphics` is still going really strong at idno interactive! (https://idno.se/swap) and while our fork has diverged a little bit I still try to keep up with your fixes :)

---

Anyway, we now have a pretty heavy compute task that runs before our main rendering. I have setup a pair of resources that the compute and render tasks ping-pong.

The compute task works on resources A, while rendering samples resources B. Next frame the compute works on resoures B while main samples from A and so on. 

I hoped the driver would be able to see this and schedule them better, but In blade on vulkan however, this gets encoded as a compute task with a full pipeline stall in the main queue and they always run sequentially and in sync. Even though the renderpass doesn't sample from the result of the compute, it still waits for it to finish.

On Apple M1 and M4, the driver succesfully sees that there is no dependency in the current frame and can thus schedule the compute at the same time as the main fragment rendering.

I don't know if we need any semaphores/`MTLSharedEvent` with cross-queue intra-frame synchronization (yet 😅), but if we every frame could submit one encoder on the main queue and one encoder on a secondary compute queue, we could still use blade's simple single submit syncpoints from the separate queue submissions to schedule this. I am not very familiar with multiple queues yet though so I really don't know.

I will be experimenting in the next week to see if I can hack together a version on my fork that just has a second `Context::submit_async()` that (if the device supports more than one queue) will submit on a secondary queue and otherwise just fallback to the main queue. I will measure and come back with the results to see if the driver is able to figure it out.

Just wanted to let you know and if you perhaps see a different approach that could solve our usecase I am all ears, and if you have any design suggestions before I potentially submit a PR.

kvark:
Closes #329

### DIFF
--- a/blade-graphics/src/gles/mod.rs
+++ b/blade-graphics/src/gles/mod.rs
@@ -459,6 +459,7 @@ impl Context {
             dual_source_blending: false,
             shader_float16: false,
             cooperative_matrix: crate::CooperativeMatrix::default(),
+            queues: vec![crate::QueueType::Main],
         }
     }
 
--- a/blade-graphics/src/gles/mod.rs
+++ b/blade-graphics/src/gles/mod.rs
@@ -438,9 +438,9 @@ pub struct PipelineContext<'a> {
     limits: &'a Limits,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
 pub struct SyncPoint {
-    fence: glow::Fence,
+    fence: Option<glow::Fence>,
 }
 //TODO: destructor
 
@@ -585,12 +585,16 @@ impl crate::traits::CommandDevice for Context {
         for frame in encoder.present_frames.drain(..) {
             self.platform.present(frame);
         }
-        SyncPoint { fence }
+        SyncPoint { fence: Some(fence) }
     }
 
     fn wait_for(&self, sp: &SyncPoint, timeout_ms: u32) -> Result<bool, crate::DeviceError> {
         use glow::HasContext as _;
 
+        let fence = match sp.fence {
+            Some(fence) => fence,
+            None => return Ok(true), // default SyncPoint is already complete
+        };
         let gl = self.lock();
         let timeout_ns = if timeout_ms == !0 {
             !0
@@ -601,7 +605,7 @@ impl crate::traits::CommandDevice for Context {
         let timeout_ns_i32 = timeout_ns.min(MAX_TIMEOUT) as i32;
 
         let status =
-            unsafe { gl.client_wait_sync(sp.fence, glow::SYNC_FLUSH_COMMANDS_BIT, timeout_ns_i32) };
+            unsafe { gl.client_wait_sync(fence, glow::SYNC_FLUSH_COMMANDS_BIT, timeout_ns_i32) };
         match status {
             glow::ALREADY_SIGNALED | glow::CONDITION_SATISFIED => Ok(true),
             glow::TIMEOUT_EXPIRED => Ok(false),
--- a/blade-graphics/src/gles/mod.rs
+++ b/blade-graphics/src/gles/mod.rs
@@ -382,6 +382,7 @@ struct TimingData {
 
 pub struct CommandEncoder {
     name: String,
+    queue_type: crate::QueueType,
     commands: Vec<Command>,
     plain_data: Vec<u8>,
     string_data: Vec<u8>,
@@ -513,6 +514,7 @@ impl crate::traits::CommandDevice for Context {
         };
         CommandEncoder {
             name: desc.name.to_string(),
+            queue_type: desc.queue,
             commands: Vec::new(),
             plain_data: Vec::new(),
             string_data: Vec::new(),
@@ -537,7 +539,7 @@ impl crate::traits::CommandDevice for Context {
         }
     }
 
-    fn submit(&self, encoder: &mut CommandEncoder) -> SyncPoint {
+    fn submit(&self, encoder: &mut CommandEncoder, _after: &[SyncPoint]) -> SyncPoint {
         use glow::HasContext as _;
 
         let fence = {
--- a/blade-graphics/src/lib.rs
+++ b/blade-graphics/src/lib.rs
@@ -162,6 +162,10 @@ pub struct ContextDesc {
     pub overlay: bool,
     /// Force selection of a specific Device ID.
     pub device_id: Option<u32>,
+    /// Enable multi-queue support (async compute and transfer).
+    /// When enabled, every `submit` call must provide explicit
+    /// synchronization via a non-empty list of sync points.
+    pub multi_queue: bool,
 }
 
 #[derive(Debug)]
@@ -870,12 +874,16 @@ pub struct ShaderDesc<'a> {
     pub naga_module: Option<naga::Module>,
 }
 
-#[derive(Clone, Debug, Default, PartialEq)]
-pub enum CommandType {
-    Transfer,
-    Compute,
+/// Type of GPU queue to submit work to.
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub enum QueueType {
+    /// Main graphics+compute+transfer queue.
     #[default]
-    General,
+    Main,
+    /// Dedicated async compute queue.
+    AsyncCompute,
+    /// Dedicated async transfer queue.
+    AsyncTransfer,
 }
 
 pub struct CommandEncoderDesc<'a> {
@@ -884,6 +892,8 @@ pub struct CommandEncoderDesc<'a> {
     /// For example, one buffer is being run on GPU while the
     /// other is being actively encoded, which makes 2.
     pub buffer_count: u32,
+    /// Queue to submit commands to.
+    pub queue: QueueType,
 }
 
 pub struct ComputePipelineDesc<'a> {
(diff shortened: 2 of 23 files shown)


blade-graphics/src/gles/mod.rs
```diff
@@ -437,9 +438,9 @@ pub struct PipelineContext<'a> {
     limits: &'a Limits,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
```
EriKWDev, line 441: SyncPoint being default helps a lot with the ergonomics, nice!

blade-graphics/src/vulkan/init.rs
```diff
@@ -287,7 +289,42 @@ fn inspect_adapter(
...
+    let queue_family_index = queue_families
+        .iter()
+        .position(|qf| qf.queue_flags.contains(vk::QueueFlags::GRAPHICS))
+        .unwrap_or(0) as u32;
+    // Find dedicated async compute queue family (compute+transfer but NOT graphics)
+    let async_compute_queue_family = if desc.multi_queue {
+        queue_families
+            .iter()
+            .position(|qf| {
+                qf.queue_flags.contains(vk::QueueFlags::COMPUTE)
```
EriKWDev, line 315: Does the `COMPUTE` flag imply ability to transfer?

EriKWDev: Interesting, yes. So if they report COMPUTE they don't have to report TRANSFER..

> All commands that are allowed on a queue that supports transfer operations are also allowed on a queue that supports either graphics or compute operations. Thus, if the capabilities of a queue family include [VK_QUEUE_GRAPHICS_BIT](https://docs.vulkan.org/spec/latest/chapters/devsandqueues.html?utm_source=chatgpt.com#VkQueueFlagBits) or [VK_QUEUE_COMPUTE_BIT](https://docs.vulkan.org/spec/latest/chapters/devsandqueues.html?utm_source=chatgpt.com#VkQueueFlagBits), then reporting the [VK_QUEUE_TRANSFER_BIT](https://docs.vulkan.org/spec/latest/chapters/devsandqueues.html?utm_source=chatgpt.com#VkQueueFlagBits) capability separately for that queue family is optional.
- https://docs.vulkan.org/spec/latest/chapters/devsandqueues.html#VkQueueFlagBits

blade-render/src/util/frame_pacer.rs
```diff
@@ -55,14 +54,14 @@ impl FramePacer {
     }
 
     pub fn end_frame(&mut self, context: &blade_graphics::Context) -> &blade_graphics::SyncPoint {
```
EriKWDev, line 56: Probably not useful for you, but we actually use a version of this framepacer and temporary resources concept in our engine too :) Though I added support for all kinds of resources as well as more than just one in flight. So I will need to add the `after: &[SyncPoint]`, here or somehow integrate it into the pacer

blade-graphics/src/lib.rs
```diff
@@ -162,6 +162,10 @@ pub struct ContextDesc {
     pub overlay: bool,
     /// Force selection of a specific Device ID.
     pub device_id: Option<u32>,
+    /// Enable multi-queue support (async compute and transfer).
+    /// When enabled, every `submit` call must provide explicit
+    /// synchronization via a non-empty list of sync points.
+    pub multi_queue: bool,
```
EriKWDev, line 168: Now we can request `multi_queue`, but it is difficult to inspect the context we get back to determine if internal async compute queue is *truly* a unique queue or just the same as main.

In the Game, if we have true async compute we will want to do our resource "Ping-Ponging", but if we don't it would be nice to only allocate one set of probe data resources and render and sample the same all the time.

So, maybe it would be nice to be able to query somehow if they are all the same queues really under the hood after Context creation, maybe `Context::get_selected_queue_id(&self, kind: QueueType) -> u32`, which would allow us to know if they are just the same.. 

Or, `Context::enumerate` could report more details in the `DeviceInfo` about the queues as well.

But maybe I am worrying about something that isn't really an issue and this is too nieche to expose.. Its not the end of the world if we have unnecessary probe datum

EriKWDev (commented):
Nice! I want to put this into the engine and see what I can to. I might need to have several encoders per frame now which is fun!

Thanks for implementing it so quickly :)

The ping-pong raytrace fits perfectly into this API. 

The other use-case I have is that one initial encoder enqueues some work, then we fire off the main encoder render task and some async compute encoder tasks that depend on that initial work being done. With this API the the async queue submission and main queue submission both can take in the initial encoder's syncpoint and the initial encoder's submit can take in the previous async compute syncpoint and main render syncpoint. I think this works too!

EriKWDev:
The particle example currently submits the compute on the async queue correctly and then render on the main queue as well, but due to the data dependency on `particle_buf` between the compute and render, they are still executed sequentially

image

So as an exercise, I will try to rewrite the particle example to have two separate buffers of particle data, one being worked on by the compute task and one being sampled from and rendered just to see if I can get the kind of parallelism I want to see.

EriKWDev:
image

Got double buffering working, but maybe I am naive in thinking this should be able to be better?

Screenshot From 2026-04-14 12-30-07_annotated

Do you think that the barrier at the end of the compute task might be preventing something? (it shouldn't? should be intra-queue only?) Or is this just the reality of the hardware?

EriKWDev:
Been working on the particle example with double buffering the particles.

Hmm, no matter what I do it seems to take roughly the same time in the end. The one outlier I got was one where the driver suddenly decided to schedule the async compute task at the very end of the vertex shader in the particle example, overlapping both vertex and fragmetn work.

I split up the compute workload into chunks and submitted it after the main submission and suddenly this happened and was much faster:

image


Other than that one exception, I get roughly the same perf from async compute as I get from removing some of the barriers internally between passes on a single queue. I did an unsafe experiment just to see and removed most of the barriers inside blade and submitted everyhitng on the main queue, and got this:

image

Also tried various permutations of splitting the render task into multiple draws of chunks of particles and splitting the compute into various chunks of compute work, but nothing seems to move the needle greatly

image

image


So it is a bit unclear from my experiments here how to get the async compute work to land in an optimal location..

<details>
  <summary>madness</summary>
  
  image
  
</details>

blade-graphics/src/lib.rs
kvark: I wanted to add this to `Capabilities` at some point but then it slipped. I'll add it.

kvark:
I just wanted to demonstrate in one of the examples *how* to specify dependencies. But you are totally right to expect the actual parallelism in the particle example. I'll look into it more.

kvark:
@EriKWDev please take another look.
I've extended the particle example with an option to toggle parallel execution on/off.
Particle-parallel
When cranking up the spawn rate, I see a strong difference between parallel ON (about 4.8ms) and OFF (about 5.7ms) on my AMD Framework 13. Radeon GPU profiler should agree with that.

EriKWDev:
Nice with getting access to what queues are available!

I'm afraid we did the same thing with the example xD But nice, now the example contains the parallelism.

The problem I tried to point out with the measurements was that my gains were very inconsistent frame to frame, and the best performance I got when the compute task overlapped with the vertex shading in this particle example.

But with this API I don't have the granularity to specify that it specifically run with the vertex and not the fragment.

In the game however, multiple encoders should be enough. We probably want the raytracing compute to always overlap with the shadowmap generation and reflection pass, and I think I can achieve it with multiple encoders and correct `after: &[]` syncpoints.

My experiment also seemed to suggest that removing the barriers internal to blade had about a similar effect to the async compute^ for this workload, but having manual barriers is perhaps undesirable. Though, I could see a way for blade to expose it with `encoder.set_use_auto_barriers(bool)` and then a manual `encoder.barrier(&self, src, dst)`. But if we need that ultimately for the game I don't know if it needs to be included in blade proper.

Also, for correctness of the simulation, I believe there has to be a buffer copy as well so that each compute task works on the resulting data of the most recent one

```rust
                if let mut transfer = self.compute_encoder.transfer("copy") {
                    let prev_compute_idx = draw_idx; // only one in flight atm

                    let a = &self.particle_systems[prev_compute_idx];
                    let b = &self.particle_systems[compute_idx];
                    // NOTE: an indirect_dispatch or at least just compute pass could copy only the alive particles..
                    transfer.copy_buffer_to_buffer(
                        a.particle_buf.at(0),
                        b.particle_buf.at(0),
                        a.particle_buf.size(),
                    );
                    transfer.copy_buffer_to_buffer(
                        a.free_list_buf.at(0),
                        b.free_list_buf.at(0),
                        a.free_list_buf.size(),
                    );
                }
```

kvark:
Yeah, I'm not sure how I feel about skipping the barriers between passes. Something like this should be very explicit about what you are expecting to run at the same time. I'll think about it.

EriKWDev:
> Yeah, I'm not sure how I feel about skipping the barriers between passes. Something like this should be very explicit about what you are expecting to run at the same time. I'll think about it.

Yeah the barriers was just an experiment on my part and not related really, though I did open #343 as we have a perhaps much more motivating example for custom barriers, or just a single little pass without an automatic barrier.

kvark:
Interesting. So in this case multi-queue would not help you. I'm actually surprised you are seeing this much cost for the barriers.

PR #340, opened 2026-04-14
metal: better wait_for implementation in case we are waiting indefinately on a SyncPoint

EriKWDev:


blade-graphics/src/metal/mod.rs
```diff
@@ -681,6 +681,26 @@ impl crate::traits::CommandDevice for Context {
 
     fn wait_for(&self, sp: &SyncPoint, timeout_ms: u32) -> Result<bool, crate::DeviceError> {
         use metal::MTLCommandBuffer as _;
+
+        if timeout_ms == !0 {
+            /*
+                NOTE: Special case if we are waiting indefinately
+            */
+            loop {
+                sp.cmd_buf.waitUntilCompleted();
```
kvark: let's just do it like this:
```rust
if timeout_ms == !0 {
  sp.cmd_buf.waitUntilCompleted();
}
// rest as usual
```

EriKWDev: I did that first xP Should've just kept it..

The only thing we avoid is `Instant::now()`, but the complexity to benefit calculation does not add up there xD

PR #342, opened 2026-04-15
vk: Use driver API version for instance creation

### LINKED ISSUE #341: vk: Instance creation requests header API version instead of driver API version
`vkCreateInstance` is called with `api_version` set to `vk::HEADER_VERSION_COMPLETE` (Vulkan 1.3.x from the ash headers):

https://github.com/kvark/blade/blob/a072374/blade-graphics/src/vulkan/init.rs#L684

On systems where the driver only supports Vulkan 1.1 or 1.2 (e.g. lavapipe on RHEL 8), this causes `ERROR_INCOMPATIBLE_DRIVER`. The [spec](https://registry.khronos.org/vulkan/specs/latest/man/html/VkApplicationInfo.html) says conformant 1.1+ implementations must not return this error, but older Mesa builds do.

`driver_api_version` is already queried via `try_enumerate_instance_version()` a few lines above and used throughout the function for extension promotion checks. Using it here instead would fix this without changing behavior on systems that already work:

```rust
.api_version(driver_api_version)
```

This matches what wgpu-hal does for instance creation.

Happy to open a PR if this looks right.

hellno:
Use the version reported by `vkEnumerateInstanceVersion` instead of `vk::HEADER_VERSION_COMPLETE` when creating the Vulkan instance.

On systems where the driver only supports Vulkan 1.1 or 1.2 (e.g. lavapipe on RHEL 8), requesting the ash header version (1.3.x) causes `ERROR_INCOMPATIBLE_DRIVER`. The [spec](https://registry.khronos.org/vulkan/specs/latest/man/html/VkApplicationInfo.html) says conformant 1.1+ implementations must not return this error, but older Mesa builds do.

`driver_api_version` is already queried via `try_enumerate_instance_version()` and used for extension promotion checks in the same function. This matches what wgpu-hal does for instance creation.

Fixes #341

### DIFF
--- a/blade-graphics/src/vulkan/init.rs
+++ b/blade-graphics/src/vulkan/init.rs
@@ -681,7 +681,7 @@ impl super::VulkanInstance {
             let app_info = vk::ApplicationInfo::default()
                 .engine_name(c"blade")
                 .engine_version(1)
-                .api_version(vk::HEADER_VERSION_COMPLETE);
+                .api_version(driver_api_version);
             let str_pointers = layers
                 .iter()
                 .chain(enabled_instance_extensions.iter())
--- a/blade-graphics/src/vulkan/init.rs
+++ b/blade-graphics/src/vulkan/init.rs
@@ -681,7 +681,7 @@ impl super::VulkanInstance {
             let app_info = vk::ApplicationInfo::default()
                 .engine_name(c"blade")
                 .engine_version(1)
-                .api_version(driver_api_version);
+                .api_version(driver_api_version.min(vk::HEADER_VERSION_COMPLETE));
             let str_pointers = layers
                 .iter()
                 .chain(enabled_instance_extensions.iter())


blade-graphics/src/vulkan/init.rs
```diff
@@ -681,7 +681,7 @@ impl super::VulkanInstance {
             let app_info = vk::ApplicationInfo::default()
                 .engine_name(c"blade")
                 .engine_version(1)
-                .api_version(vk::HEADER_VERSION_COMPLETE);
+                .api_version(driver_api_version);
```
kvark: we should take minimum of ours and theirs, not just theirs

kvark (commented):
Thanks for the PR!

blade-graphics/src/vulkan/init.rs
hellno: oh yeah, changed it!

hellno:
changed it to use the min version between the two

kvark (approved):

kvark:
Ah, I should have included this in 0.8.3. release

PR #353, opened 2026-05-12
WebGL2/wasm/Android: GLES buffer, texture, and fence fixes; GLES/Vulkan + egui robustness

ohsalmeron:
### Problem
Running Blade on **WebGL2 / wasm** hit spec and browser constraints: the same `WebGLBuffer` cannot be bound as both `ARRAY_BUFFER` and `ELEMENT_ARRAY_BUFFER`; `clientWaitSync` cannot block on the main thread; texture unit / unpack state could produce wrong sampling or extra copies (e.g. Firefox). On **GLES**, integer texture formats and uniform-block reflection needed fixes, plus explicit buffer sync for non-`bufferStorage` paths. **blade-egui** needed a proper vertex layout/fetch path, split vertex/index belts, and deferred texture destruction after submit. **Vulkan** debug utils were assumed present.

### What changed
- **wasm / WebGL2:** split vertex/index `BufferBelt`s with explicit `BufferDesc.bind_point` (no dual-buffer overhead on every buffer); `client_wait_sync` timeout `0` on WebGL (runtime `is_webgl` + `MAX_CLIENT_WAIT_TIMEOUT_WEBGL` query); texture target unbind hygiene; `UNPACK_ROW_LENGTH` only when stride differs from width (restored to 0 after upload); texture units deduplicated when Naga lists the same GLSL sampler name for texture and sampler bindings.
- **GLES:** `R32Uint` uses `RED_INTEGER` external format; uniform blocks resolved via Naga reflection name + WGSL struct type name only (hard error on miss — no `type_N` or single-block guessing); `sync_buffer(buffer, offset, size)` on the trait (replaces GLES-only `sync_buffer_range`); EGL surfaceless init keeps `.unwrap()` as intended.
- **blade-egui:** vertex input layout + fetch; dual fragment entries — `fs_main` outputs linear (default `ColorSpace::Linear` convention), `fs_main_srgb` gamma passthrough for plain UNORM surfaces (selected when swapchain format is `Rgba8Unorm`, e.g. WebGL canvas blit); fragment stage does not read uniform buffers (avoids WebGL link errors); textures-to-free queue + `sync()` after submit.
- **Web canvas:** Blade does **not** resize the embedder's DOM canvas; the application sets `#blade` backing-store pixels before surface reconfigure.
- **Vulkan:** enable and use `VK_EXT_debug_utils` only when the extension exists; guard naming calls.

Rebased onto latest `kvark/main` (includes `manual_barriers` on `CommandEncoderDesc`, Naga update).

### Testing
- [x] `wasm32-unknown-unknown` — Chrome / Firefox WebGL2 (egui + game map/sprites)
- [x] Native Vulkan smoke (Linear swapchain, egui + map color correct)
- [ ] Android GLES (older Mali) — not re-verified in this pass

### Follow-ups for review
- Uniform block lookup: if a driver still fails with named WGSL structs, I will attach a minimal Naga→GLSL repro rather than reintroduce name guessing.

### DIFF
--- a/blade-egui/shader.wgsl
+++ b/blade-egui/shader.wgsl
@@ -44,9 +44,8 @@ var r_sampler: sampler;
 
 @fragment
 fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
-    //Note: we always assume rendering to linear color space,
-    // but Egui wants to blend in gamma space, see
-    // https://github.com/emilk/egui/pull/2071
+    //Note: we output sRGB directly because our swapchain is configured as Srgb 
+    // which gives us a Unorm format (no hardware conversion).
     let blended = in.color * textureSample(r_texture, r_sampler, in.tex_coord);
-    return vec4f(linear_from_gamma(blended.xyz), blended.a);
+    return vec4f(blended.xyz, blended.a);
 }
--- a/blade-egui/shader.wgsl
+++ b/blade-egui/shader.wgsl
@@ -6,8 +6,7 @@ struct VertexOutput {
 
 struct Uniforms {
     screen_size: vec2<f32>,
-    convert_to_linear: f32,
-    padding: f32,
+    padding: vec2<f32>,
 };
 var<uniform> r_uniforms: Uniforms;
 
@@ -42,10 +41,21 @@ fn vs_main(input: VertexInput) -> VertexOutput {
 var r_texture: texture_2d<f32>;
 var r_sampler: sampler;
 
+// Egui blends in gamma space, see https://github.com/emilk/egui/pull/2071
+fn blended_color(in: VertexOutput) -> vec4<f32> {
+    return in.color * textureSample(r_texture, r_sampler, in.tex_coord);
+}
+
 @fragment
 fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
-    //Note: we output sRGB directly because our swapchain is configured as Srgb 
-    // which gives us a Unorm format (no hardware conversion).
-    let blended = in.color * textureSample(r_texture, r_sampler, in.tex_coord);
+    // ColorSpace::Linear swapchain (Vulkan/Metal/EGL sRGB storage).
+    let blended = blended_color(in);
+    return vec4f(linear_from_gamma(blended.xyz), blended.a);
+}
+
+@fragment
+fn fs_main_srgb(in: VertexOutput) -> @location(0) vec4<f32> {
+    // Plain UNORM canvas (WebGL blit to HTML canvas): output gamma-space directly.
+    let blended = blended_color(in);
     return vec4f(blended.xyz, blended.a);
 }
--- a/blade-egui/shader.wgsl
+++ b/blade-egui/shader.wgsl
@@ -12,33 +12,27 @@ struct Uniforms {
 var<uniform> r_uniforms: Uniforms;
 
 //Note: avoiding `vec2<f32>` in order to keep the scalar alignment
-struct Vertex {
-    pos_x: f32,
-    pos_y: f32,
-    tex_coord_x: f32,
-    tex_coord_y: f32,
-    color: u32,
+struct VertexInput {
+    a_pos: vec2<f32>,
+    a_tex_coord: vec2<f32>,
+    a_color: u32,
 }
-var<storage, read> r_vertex_data: array<Vertex>;
 
 fn linear_from_gamma(srgb: vec3<f32>) -> vec3<f32> {
-    let cutoff = srgb < vec3<f32>(0.04045);
     let lower = srgb / vec3<f32>(12.92);
     let higher = pow((srgb + vec3<f32>(0.055)) / vec3<f32>(1.055), vec3<f32>(2.4));
-    return select(higher, lower, cutoff);
+    let is_higher = step(vec3<f32>(0.04045), srgb);
+    return mix(lower, higher, is_higher);
 }
 
 @vertex
-fn vs_main(
-    @builtin(vertex_index) v_index: u32,
-) -> VertexOutput {
-    let input = r_vertex_data[v_index];
+fn vs_main(input: VertexInput) -> VertexOutput {
     var out: VertexOutput;
-    out.tex_coord = vec2<f32>(input.tex_coord_x, input.tex_coord_y);
-    out.color = unpack4x8unorm(input.color);
+    out.tex_coord = input.a_tex_coord;
+    out.color = unpack4x8unorm(input.a_color);
     out.position = vec4<f32>(
-        2.0 * input.pos_x / r_uniforms.screen_size.x - 1.0,
-        1.0 - 2.0 * input.pos_y / r_uniforms.screen_size.y,
+        2.0 * input.a_pos.x / r_uniforms.screen_size.x - 1.0,
+        1.0 - 2.0 * input.a_pos.y / r_uniforms.screen_size.y,
         0.0,
         1.0,
     );
--- a/blade-graphics/src/gles/command.rs
+++ b/blade-graphics/src/gles/command.rs
@@ -109,7 +109,11 @@ impl super::CommandEncoder {
 
     pub fn barrier(&mut self) {}
 
-    fn pass<P>(&mut self, kind: super::PassKind, target_size: [u16; 2]) -> super::PassEncoder<'_, P> {
+    fn pass<P>(
+        &mut self,
+        kind: super::PassKind,
+        target_size: [u16; 2],
+    ) -> super::PassEncoder<'_, P> {
         super::PassEncoder {
             commands: &mut self.commands,
             plain_data: &mut self.plain_data,
@@ -319,13 +323,16 @@ impl super::PassEncoder<'_, super::ComputePipeline> {
 impl crate::traits::RenderEncoder for super::PassEncoder<'_, super::RenderPipeline> {
     fn set_scissor_rect(&mut self, rect: &crate::ScissorRect) {
         // Invert Y axis for OpenGL's bottom-left window coordinates
-        let y = (self.target_size[1] as i32).saturating_sub(rect.y).saturating_sub(rect.h as i32);
-        self.commands.push(super::Command::SetScissor(crate::ScissorRect {
-            x: rect.x,
-            y: y.max(0),
-            w: rect.w,
-            h: rect.h,
-        }));
+        let y = (self.target_size[1] as i32)
+            .saturating_sub(rect.y)
+            .saturating_sub(rect.h as i32);
+        self.commands
+            .push(super::Command::SetScissor(crate::ScissorRect {
+                x: rect.x,
+                y: y.max(0),
+                w: rect.w,
+                h: rect.h,
+            }));
     }
 
     fn set_viewport(&mut self, viewport: &crate::Viewport) {
@@ -669,7 +676,10 @@ impl super::Command {
                     base_vertex,
                     instance_count,
                 } => {
+                    #[cfg(not(target_arch = "wasm32"))]
                     gl.bind_buffer(glow::ELEMENT_ARRAY_BUFFER, Some(index_buf.raw));
+                    #[cfg(target_arch = "wasm32")]
+                    gl.bind_buffer(glow::ELEMENT_ARRAY_BUFFER, Some(index_buf.raw_index));
                     match (base_vertex, instance_count) {
                         (0, 1) => gl.draw_elements(
                             topology,
@@ -791,7 +801,11 @@ impl super::Command {
                     let row_texels =
                         bytes_per_row / block_info.size as u32 * block_info.dimensions.0 as u32;
                     gl.pixel_store_i32(glow::UNPACK_ALIGNMENT, 1);
-                    gl.pixel_store_i32(glow::UNPACK_ROW_LENGTH, row_texels as i32);
+                    // Only set row length when stride differs from width;
+                    // redundant values trigger Firefox's slow defensive copy path.
+                    if row_texels != size.width {
+                        gl.pixel_store_i32(glow::UNPACK_ROW_LENGTH, row_texels as i32);
+                    }
                     gl.bind_buffer(glow::PIXEL_UNPACK_BUFFER, Some(src.raw));
                     gl.bind_texture(dst.target, Some(dst.raw));
                     let unpack_data = glow::PixelUnpackData::BufferOffset(src.offset as u32);
@@ -861,6 +875,7 @@ impl super::Command {
                         _ => unreachable!(),
                     }
                     gl.bind_buffer(glow::PIXEL_UNPACK_BUFFER, None);
+                    gl.pixel_store_i32(glow::UNPACK_ROW_LENGTH, 0);
                 }
                 Self::CopyTextureToBuffer {
                     ref src,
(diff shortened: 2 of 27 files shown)


blade-graphics/src/gles/command.rs
```diff
@@ -1204,6 +1227,21 @@ impl super::Command {
                     target,
                 } => {
                     gl.active_texture(glow::TEXTURE0 + slot);
+                    #[cfg(target_arch = "wasm32")]
```
kvark: can we turn this into a normal if condition, e.g. `if cfg!(target_arch = "wasm32") {...}`

blade-graphics/src/gles/egl.rs
```diff
@@ -326,15 +326,13 @@ impl super::Context {
...
                         egl::ATTRIB_NONE,
                     ];
-                    Some(
-                        egl1_5
-                            .get_platform_display(
-                                EGL_PLATFORM_ANGLE_ANGLE,
-                                ptr::null_mut(),
-                                &display_attributes,
-                            )
-                            .unwrap(),
```
kvark, line 336: this was actually intentional: we don't expect it to fail, and we do want to see the error, not just skip over it

blade-graphics/src/gles/egl.rs
```diff
@@ -350,14 +348,21 @@ impl super::Context {
...
-                        .unwrap();
-                    (d, None)
+                    let d = egl1_5.get_platform_display(
+                        EGL_PLATFORM_SURFACELESS_MESA,
+                        ptr::null_mut(),
+                        &[egl::ATTRIB_NONE],
+                    );
+                    match d {
+                        Ok(display) => (display, None),
+                        Err(e) => {
```
kvark: why would this fail?

blade-graphics/src/gles/mod.rs
```diff
@@ -586,17 +594,26 @@ impl crate::traits::CommandDevice for Context {
...
         let gl = self.lock();
-        let timeout_ns = if timeout_ms == !0 {
-            !0
-        } else {
-            timeout_ms as u64 * 1_000_000
+        // WebGL2: MAX_CLIENT_WAIT_TIMEOUT_WEBGL is 0, blocking is forbidden.
+        #[cfg(target_arch = "wasm32")]
+        let timeout_ns_i32 = 0i32;
+        #[cfg(not(target_arch = "wasm32"))]
+        let timeout_ns_i32 = {
```
kvark: let's prefer run-time checks over compile time: more compilation code coverage

blade-graphics/src/gles/mod.rs
```diff
@@ -641,9 +658,9 @@ fn describe_texture_format(format: crate::TextureFormat) -> FormatInfo {
         Tf::R32Float => (glow::R32F, glow::RED, glow::FLOAT),
         Tf::Rg32Float => (glow::RG32F, glow::RG, glow::FLOAT),
         Tf::Rgba32Float => (glow::RGBA32F, glow::RGBA, glow::FLOAT),
-        Tf::R32Uint => (glow::R32UI, glow::RED, glow::UNSIGNED_INT),
-        Tf::Rg32Uint => (glow::RG32UI, glow::RG, glow::UNSIGNED_INT),
-        Tf::Rgba32Uint => (glow::RGBA32UI, glow::RGBA, glow::UNSIGNED_INT),
+        Tf::R32Uint => (glow::R32UI, glow::RED_INTEGER, glow::UNSIGNED_INT),
```
kvark, line 661: nice catch!

blade-graphics/src/gles/pipeline.rs
```diff
@@ -146,7 +146,12 @@ impl super::Context {
 
                 for mapping in attribute_mappings {
                     let vf = &vertex_fetch_states[mapping.buffer_index];
-                    let (_, attrib) = vf.layout.attributes[mapping.attribute_index];
+                    let (attrib_name, attrib) = vf.layout.attributes[mapping.attribute_index];
+
+                    let location = attributes.len() as u32;
+                    gl.bind_attrib_location(program, location, attrib_name);
+                    gl.bind_attrib_location(program, location, &format!("e_{}", attrib_name));
```
kvark: what's happening here? why are we binding it twice?

blade-graphics/src/gles/pipeline.rs
```diff
@@ -243,9 +250,15 @@ impl super::Context {
                                 if let Some(ref location) =
                                     gl.get_uniform_location(program, glsl_name)
                                 {
-                                    let mut slots = [0i32];
-                                    gl.get_uniform_i32(program, location, &mut slots);
-                                    targets.push(slots[0] as u32);
+                                    let slot = *texture_slots
+                                        .entry(glsl_name.to_string())
```
kvark, line 255: when would this happen? is this about using the same texture in multiple stages?

blade-graphics/src/gles/pipeline.rs
```diff
@@ -268,8 +281,28 @@ impl super::Context {
                                 unimplemented!()
                             }
                             crate::ShaderBinding::Plain { size } => {
-                                if let Some(index) = gl.get_uniform_block_index(program, glsl_name)
-                                {
+                                let mut index_opt = gl.get_uniform_block_index(program, glsl_name);
+                                if index_opt.is_none() {
+                                    if let Some(ref type_name) = sf.shader.module.types[var.ty].name
+                                    {
```
kvark, line 289: nit: not formatted?

blade-graphics/src/gles/pipeline.rs
```diff
@@ -268,8 +281,28 @@ impl super::Context {
...
+                                        index_opt = gl.get_uniform_block_index(program, type_name);
+                                    }
+                                }
+                                if index_opt.is_none() {
+                                    let type_name = format!("type_{}", var.ty.index());
```
kvark: why so many paths to bind?

blade-graphics/src/gles/mod.rs
```diff
@@ -48,6 +49,8 @@ pub struct Surface {
 #[derive(Clone, Copy, Debug, Hash, PartialEq)]
 pub struct Buffer {
     raw: glow::Buffer,
+    #[cfg(target_arch = "wasm32")]
+    raw_index: glow::Buffer,
```
kvark: This is really unfortunate. It makes all of the buffers on WebGL less efficient, just because there is one quirk in the spec.
What do you think of a counter-proposal: we just assert that the user isn't hitting this path on webgl but keep the implementation clean? that would lift the burden to the user.

kvark (commented):
Looks to be very well thought, thank you for filing!

ohsalmeron:
Hi kvark, thanks for the review! 

Let's keep the buffer implementation clean, and I want to make sure we do this the right way.

I've addressed all the minor comments:
-Switched compile-time #[cfg] blocks to run-time cfg! for WASM to ensure full compilation coverage.
-Updated the EGL fallback to use .unwrap() cleanly as intended.
-Fixed formatting nits. (Note on the e_ prefix and uniform block variants: Naga GLSL transpilation sometimes adds these depending on the config. Binding both is a defensive check to ensure robust matching across different browsers/drivers).

Regarding the WebGL2 index buffer constraint: I  agree we shouldn't add overhead or complexity to the native Vulkan/Metal paths. My challenge is that because WebGL2 strictly forbids sharing the same buffer for ARRAY and ELEMENT_ARRAY, enforcing this on the user would force me to heavily restructure my engine's entire pipeline just to support WebGL/Firefox.

Could I get a bit of time to dig deeper into the internals here? I want to explore if we can design a cleaner abstraction that satisfies the restrictive WebGL spec without punishing native buffer efficiency at all. 

I'm fully committed to getting this right for the upstream!

blade-graphics/src/gles/resource.rs
```diff
@@ -61,23 +61,29 @@ impl crate::traits::ResourceDevice for super::Context {
             crate::Memory::External(_) => unimplemented!(),
         };
 
+        let target = if desc.name.contains("index") {
```
kvark, line 64: The name check is awkward here. All names should be non-semantical.
But really, do we have to check for this at all? WebGL forbids binding the buffer to 2 different bind points simultaneously. This code here is just for data initialization. We can unbind the buffer freely at the end of this function, and it's not bound anywhere else, so should be totally valid to always use `ARRAY_BUFFER` like the original code did.
Basically, I wonder if everything will be correct if we just not introduce the name at all. All we should need is that split between the index belt and the vertex belt.

blade-graphics/src/gles/resource.rs
```diff
@@ -204,6 +227,9 @@ impl crate::traits::ResourceDevice for super::Context {
                 // assigning the storage.
                 gl.tex_parameter_i32(target, glow::TEXTURE_MIN_FILTER, glow::NEAREST as i32);
                 gl.tex_parameter_i32(target, glow::TEXTURE_MAG_FILTER, glow::NEAREST as i32);
+                gl.tex_parameter_i32(target, glow::TEXTURE_WRAP_S, glow::CLAMP_TO_EDGE as i32);
```
kvark: Is this not defined by default?

blade-graphics/src/gles/web.rs
```diff
@@ -126,6 +126,17 @@ impl super::Context {
...
         //Note: this code can be shared with EGL
+        let canvas = web_sys::window()
+            .unwrap()
+            .document()
+            .unwrap()
+            .get_element_by_id("blade")
+            .unwrap()
+            .dyn_into::<web_sys::HtmlCanvasElement>()
+            .unwrap();
+        canvas.set_width(config.size.width);
```
kvark: This looks weird to me. It would be analogous if on native we'd try to resize the window here. The canvas is providing us with a surface, but we should not own it here.

blade-graphics/src/gles/resource.rs
```diff
@@ -102,15 +109,31 @@ impl crate::traits::ResourceDevice for super::Context {
...
                 let data = slice::from_raw_parts(buffer.data, buffer.size as usize);
-                gl.bind_buffer(glow::ARRAY_BUFFER, Some(buffer.raw));
-                gl.buffer_sub_data_u8_slice(glow::ARRAY_BUFFER, 0, data);
+                gl.bind_buffer(buffer.target, Some(buffer.raw));
+                gl.buffer_sub_data_u8_slice(buffer.target, 0, data);
+            }
+        }
+    }
+
+    fn sync_buffer_range(&self, buffer: super::Buffer, offset: u64, size: u64) {
```
kvark: should we just change the signature of the original `sync_buffer` method? It's not relevant to other backends anyway, and the user can get the size of the buffer by doing `buffer.size()` shall they need to sync all of it.

blade-graphics/src/gles/pipeline.rs
```diff
@@ -268,14 +282,36 @@ impl super::Context {
                                 unimplemented!()
                             }
                             crate::ShaderBinding::Plain { size } => {
-                                if let Some(index) = gl.get_uniform_block_index(program, glsl_name)
-                                {
+                                // Naga may emit the uniform block name as the variable name,
```
kvark: this looks suss to me. Naga should always expose to us explicitly what the generated names are, given that the client may need to know them. We shouldn't ever guess.

blade-graphics/src/gles/pipeline.rs
```diff
@@ -146,7 +146,13 @@ impl super::Context {
 
                 for mapping in attribute_mappings {
                     let vf = &vertex_fetch_states[mapping.buffer_index];
-                    let (_, attrib) = vf.layout.attributes[mapping.attribute_index];
+                    let (attrib_name, attrib) = vf.layout.attributes[mapping.attribute_index];
+
+                    let location = attributes.len() as u32;
+                    gl.bind_attrib_location(program, location, attrib_name);
+                    // Naga's GLSL backend prefixes entry point arguments with 'e_'
```
kvark: Similarly suspicious. Naga's "e_" may be an implementation detail, not a part of the contract.

blade-graphics/src/gles/mod.rs
```diff
@@ -586,17 +588,25 @@ impl crate::traits::CommandDevice for Context {
...
         } else {
-            timeout_ms as u64 * 1_000_000
+            let timeout_ns = if timeout_ms == !0 {
+                !0
+            } else {
+                timeout_ms as u64 * 1_000_000
+            };
+            timeout_ns.min(MAX_TIMEOUT) as i32
         };
-        //TODO: https://github.com/grovesNL/glow/issues/287
```
kvark, line 598: this comment is still useful, shouldn't remove it

blade-graphics/src/gles/mod.rs
```diff
@@ -586,17 +588,25 @@ impl crate::traits::CommandDevice for Context {
...
 
         let gl = self.lock();
-        let timeout_ns = if timeout_ms == !0 {
-            !0
+        // WebGL2: MAX_CLIENT_WAIT_TIMEOUT_WEBGL is 0, blocking is forbidden.
+        let timeout_ns_i32 = if cfg!(target_arch = "wasm32") {
+            0i32
         } else {
-            timeout_ms as u64 * 1_000_000
+            let timeout_ns = if timeout_ms == !0 {
```
kvark: is this a no-op today? we are still doing `min(MAX_TIMEOUT)` after that, so `1_000_000 * u32::MAX` is behaving the same as `!0`

blade-egui/shader.wgsl
```diff
@@ -50,9 +44,8 @@ var r_sampler: sampler;
 
 @fragment
 fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
-    //Note: we always assume rendering to linear color space,
-    // but Egui wants to blend in gamma space, see
-    // https://github.com/emilk/egui/pull/2071
+    //Note: we output sRGB directly because our swapchain is configured as Srgb
```
kvark: This looks wrong to me, and it's changing behavior for everybody, not just WebGL.
The user decides how its fragment colors are going to be presented here - https://github.com/kvark/blade/blob/0d693426253fb2aaea0d842bf76bf13e2773a411/blade-graphics/src/lib.rs#L1356-L1362

The color space is `Linear` by default in Blade - this is a convention, by which the user is expected to produce linear colors from their shaders.

ohsalmeron: Hi @kvark,

I have updated the PR and addressed most if not all of your feedback:

1. **Reverted GLES `bind_point` API Change**: 
  Since data initialization and sync only bind the buffer temporarily, we do not need to carry a GLES-specific `bind_point` on the public `BufferDesc` API. I have reverted it entirely.

- Inside `sync_buffer` and `create_buffer`, we now always bind to `glow::ARRAY_BUFFER` and **unbind it immediately afterwards** (`gl.bind_buffer(glow::ARRAY_BUFFER, None)`). This ensures we never leave a buffer bound to `ARRAY_BUFFER` after synchronization.

The WebGL2 conflict is solved by keeping the split between `vertex_belt` and `index_belt` in `blade-egui` so that the vertex and index data reside in separate underlying buffers during draw calls.

2. **WASM Runtime checks**:
   Switched the WASM compile-time checks to runtime checks (`self.is_webgl` and `cfg!(target_arch = "wasm32")`) for compilation coverage.

3. **EGL surfaceless unwrap**:
   Retained the `.unwrap()` on `get_platform_display` Mesa platform call as you intended, ensuring EGL errors are thrown explicitly rather than skipped.

4. **Naga Reflection & Guessing**:
   Removed the guessing paths (including the `e_` attribute prefix and Single-Block fallback guess). We now look up uniform blocks via Naga reflection name first, then by the WGSL struct type name. If both fail, we panic with a descriptive error listing all active uniform blocks found on the device.

5. **Egui Shader / sRGB Output**:
   Split the fragment shader entry points into `fs_main` (gamma-to-linear conversion for sRGB swapchains like Vulkan/Metal/EGL) and `fs_main_srgb` (gamma passthrough for plain UNORM canvas blits in WebGL), preserving the default `ColorSpace::Linear` conventions for other backends.

6. **Canvas Resizing**:
   Removed the DOM canvas resizing from the GLES backend initialization, leaving canvas backing-store sizing to be managed strictly by the embedder.


The primary driver for these fixes was getting, Shadows of War running cleanly on WebGL2/wasm and Android. 

If you are interested in seeing this PR in action, you can check out the game's repository here:
**https://github.com/ohsalmeron/shadows-of-war**

You can run it locally with:
* `./sow` (for the native client Vulkan or Metal)
* `./sow l` or `./sow local` (to build and serve the local wasm/WebGL2 build)

kvark:
@ohsalmeron I've got a number of WebGL fixes recently to get https://github.com/kvark/vandals-and-heroes online (you are welcome to check it out!). Sorry about not tracking the state of this PR. If you find time to rebase it and see what's left to fix, I'll be happy to review again.

ohsalmeron:
I'll take a look at it tomorrow, I'm also testing these new threejs games to be ported into blade.

Thanks for circling back

PR #366, opened 2026-08-12
Fix ReSTIR target reuse for external denoisers

kvark:
## What changed

- Re-evaluate visibility and the target function at the receiving surface for standard ReSTIR reservoir reuse instead of selecting with a stale source-surface score.
- Expose a one-shot `RayTracer::fill_gbuffer` pass so a sparse canonical path trace can feed an external denoiser/upscaler without running ReSTIR.
- Exercise the corrected non-pairwise reuse path in the PBR image reference test, update its golden image at Blade's normal strict threshold, and add a canonical-mode G-buffer regression.
- Authenticate the Windows Vulkan/LavaPipe installer so shared hosted-runner API rate limits cannot fail CI before tests start.

## Why

A reused reservoir could retain a non-zero source target at a recipient where the sample was occluded or the BRDF was black. It could then be selected and shade to zero, producing persistent dark patches. Separately, canonical path tracing did not populate Blade's G-buffer, which blocked path-tracer-first Ommatidium integration.

The remaining difference between ReSTIR and canonical illumination on occluded faces is expected: Blade's real-time estimator is direct-only while the canonical tracer includes indirect bounces. This PR does not disguise that difference with ambient illumination.

## Validation

- `cargo test --test gpu_examples snapshot_pbr_ray_trace -- --ignored --nocapture --test-threads=1` (SSIM 0.9999 against the corrected golden at Blade's standard 0.995 threshold)
- `cargo test --test gpu_examples snapshot_pbr_canonical -- --ignored --nocapture --test-threads=1` (SSIM 0.9998; ReSTIR mean difference 3.96/255)
- `cargo test --test gpu_examples gbuffer_views_describe_the_rendered_frame -- --ignored --nocapture --test-threads=1`
- `cargo clippy --workspace --all-targets` (passes with pre-existing warnings)
- GitHub Actions final Linux/Vulkan, Windows/Vulkan, macOS/Metal, MSRV, clippy, and platform-build matrix: all green


blade-render/code/ray-trace.wgsl
```diff
@@ -466,8 +466,22 @@ fn compute_restir(surface: Surface, pixel: vec2<i32>, rng: ptr<function, RandomS
...
         } else {
-            let radiance = evaluate_reflected_light(surface, neighbor.light_index, neighbor.light_uv);
-            other = unpack_reservoir(neighbor, max_confidence, radiance);
+            // A reservoir stores the target score at the surface that created
+            // it. Reusing that score at this surface gives a sample non-zero
+            // selection weight even when it is occluded or its BRDF is black
+            // here; if selected, that sample then shades to zero and creates
+            // the persistent dark patches temporal reuse used to exhibit.
+            // Re-evaluate both visibility and the target at the recipient.
+            let evaluated = estimate_target_score_with_occlusion(
```
kvark, line 475: unfortunate that we can't use `unpack_reservoir` here

PR #369, opened 2026-08-15
Integration/blade volume latest

kvark:


blade-graphics/src/vulkan/resource.rs
```diff
@@ -833,3 +862,17 @@ fn fetch_external_source(
         _ => None,
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn device_address_buffers_request_matching_memory_flags() {
```
kvark: no need to have this test - remove it

blade-graphics/src/vulkan/resource.rs
```diff
@@ -491,6 +514,7 @@ impl crate::traits::ResourceDevice for super::Context {
             desc.external
                 .map_or(crate::Memory::Device, crate::Memory::External),
             desc.name,
+            false,
```
kvark: is this beneficial in any way? is there a cost to requesting the device buffer address for the memory backing an image?

blade-graphics/src/vulkan/resource.rs
```diff
@@ -16,9 +24,10 @@ impl super::Context {
         requirements: vk::MemoryRequirements,
         memory: crate::Memory,
         name: &str,
+        needs_device_address: bool,
```
kvark: I'm not yet convinced that it's worth the code complexity to make this change

blade-graphics/src/vulkan/descriptor.rs
```diff
@@ -98,34 +187,102 @@ impl super::Device {
...
-                .reset_descriptor_pool(pool.sub_pools[0], vk::DescriptorPoolResetFlags::empty())
+                .reset_descriptor_pool(sub_pool.raw, vk::DescriptorPoolResetFlags::empty())
                 .unwrap();
         }
+        sub_pool.allocated_sets = 0;
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
kvark: I don't think we need these tests

blade-graphics/src/vulkan/descriptor.rs
```diff
@@ -98,34 +187,102 @@ impl super::Device {
         let descriptor_set_layouts = [layout.raw];
 
         loop {
+            let needs_larger_pool = {
+                let sub_pool = &pool.sub_pools[0];
+                sub_pool.allocated_sets == sub_pool.max_sets
```
kvark: not sure I understand. We used to just try allocating and see what the driver tells us. Was this incorrect for some reason? why do we need to count it on our side?

Cargo.toml
```diff
@@ -35,7 +35,7 @@ glam = { version = "0.30", features = ["mint"] }
 gltf = { version = "1.1", default-features = false }
 log = "0.4"
 mint = "0.5"
-naga = { git = "https://github.com/gfx-rs/wgpu", rev = "78b4dfd0cb1ef5ed208fefe80ae2f855db3609bb", features = ["wgsl-in", "termcolor"] }
+naga = { git = "https://github.com/kvark/wgpu", rev = "cefd48f6cff61d64abe1cdb4002c4e39bfc2c728", features = ["wgsl-in", "termcolor"] }
```
kvark: let's remove this commit, keep using upstream wgpu instead

PR #376, opened 2026-08-24
Add skeletal animation support

kvark:


blade-render/code/raster.wgsl
```diff
@@ -64,10 +71,19 @@ var shadow_tex: texture_depth_2d;
 
 var<uniform> shadow_frame_params: ShadowFrameParams;
 var<uniform> shadow_draw_params: ShadowDrawParams;
+var<uniform> shadow_skinning_params: SkinningParams;
+
+fn skin_matrix(input: Vertex, params: SkinningParams) -> mat4x4<f32> {
```
kvark: I wonder how it works under the hood to have an argument of such a gigantic size

blade-render/code/raster.wgsl
```diff
@@ -86,11 +102,14 @@ fn quat_rotate(q: vec4<f32>, v: vec3<f32>) -> vec3<f32> {
 @vertex
 fn raster_vs(input: Vertex) -> VertexOutput {
     var out: VertexOutput;
-    let pos_world = draw_params.model * vec4<f32>(input.position, 1.0);
+    let skin = skin_matrix(input, skinning_params);
```
kvark: this is unconditional? some meshes wouldn't even have weights and joints. Wouldn't it make sense to have it under a different branch arm?

blade-render/src/model/animation.rs
```diff
@@ -0,0 +1,570 @@
...
+    pub duration: f32,
+    pub channels: Vec<AnimationChannel>,
+}
+
+/// Bind transform and hierarchy link for one model node.
+#[derive(Clone, Copy, Debug)]
+pub struct SkeletonNode {
+    pub parent_index: u32,
+    pub translation: [f32; 3],
+    pub rotation: [f32; 4],
```
kvark: any reason this isn't using `mint` or `glam` types?

blade-render/src/model/animation.rs
```diff
@@ -0,0 +1,570 @@
+use std::borrow::Cow;
+
+pub(crate) const MAX_JOINTS_PER_DRAW: usize = 64;
+const NO_INDEX: u32 = u32::MAX;
+
+/// Animation keyframes loaded with a model but evaluated outside the renderer.
+#[derive(Clone, Debug)]
+pub struct AnimationClip {
```
kvark: I wonder if there is a clean way to separate the concerns here. Ideally, the renderer has nothing to do with animation. The renderer would its own representation of a model, physics would have its own, animation would have its own. And all of those components would merge together in an entity concept of blade-engine itself.

blade-engine/src/lib.rs
```diff
@@ -542,22 +547,33 @@ impl Engine {
...
+                render_object.prev_transform = gpu::Transform {
+                    x: mp.column(0).into(),
+                    y: mp.column(1).into(),
+                    z: mp.column(2).into(),
+                };
+                render_object.color_tint = object.color_tint;
+                if let Some(pose) = previous_pose {
+                    render_object.set_pose(pose);
+                }
+                if let Some(pose) = current_pose {
```
kvark: not sure what's happening here. We are calling `render_object.set_pose` twice in a row?

blade-render/code/raster.wgsl
kvark: Fixed in dba592d: the skinning functions now read the joint palette directly from their stage-specific uniform globals, so the only function argument is the vertex.

blade-render/code/raster.wgsl
kvark: Fixed in dba592d: raster and shadow rendering now have separate rigid and skinned vertex entry points/pipelines, selected from geometry.skin_index. Rigid meshes do not execute or reference skinning.

blade-render/src/model/animation.rs
kvark: Fixed in dba592d: runtime nodes and poses now use glam Vec3, Quat, and Mat4. Plain arrays remain only in serialized Flat asset data and GPU-facing structures.

blade-render/src/model/animation.rs
kvark: Refactored in dba592d: blade-engine now owns a separate AnimationModel asset/baker and AnimationPlayer evaluation. blade-render keeps only render-side skin bindings and receives a fully evaluated Pose through Object::set_pose; the engine entity joins animation, render, and physics handles.

blade-engine/src/lib.rs
kvark: Fixed in dba592d: engine render objects now persist across frames and set_pose is called exactly once with the current evaluated pose. Object::set_pose internally rolls current pose to previous pose for motion vectors, including across clip switches.

kvark:
BLAS lifecycle follow-up in dba592d:

- At scene build N, ownership of the former current TLAS plus all animated BLASes/backing buffers moves into frame N FrameResources. The ray tracer keeps only a non-owning handle while frame N uses that scene as previous-frame geometry.
- FramePacer destroys those resources only after the frame N fence signals. The ray tracer owns only the newly built current generation, so there is no extra historical generation or double-destroy path.
- Added an ignored GPU stress test with 64 continuously posed objects over 32 frames and abrupt pose-pattern switches every four frames. It asserts exactly one retired TLAS plus 64 retired BLASes are fence-owned; late-frame device memory remained flat at 56.4 MiB locally.

Also reproduced and fixed the Android CI failure (procedural Vertex literals now initialize joints/weights). The exact Android check, workspace/all-feature tests, MSRV, WebAssembly, Clippy/GLES, shader tests, raster GPU snapshot, and the new ray-tracing stress test all pass locally.

kvark:
Follow-up in 4da0bf7:

- Fixed pose history semantics: Object::set_pose now changes only the current pose. The ray tracer keys history by a stable private object ID, freezes the pose used by build_scene, and promotes it to previous-frame history only after RayTracer::render records that scene. Multiple pose assignments or unrendered scene builds therefore cannot replace the pose corresponding to the previous pixels.
- Replaced skinned: bool pipeline construction with a Variant::{Rigid, Skinned} enum.
- Skin matrices are now affine mat4x3 values. They remain 64 bytes each in a uniform because every vec3 column has 16-byte alignment.
- Positions use the affine 4x3 transform. Tangents use its 3x3 linear part. Normals use the scale-correct inverse-transpose direction via cofactors (WGSL has no matrix inverse builtin), including determinant sign for mirrored transforms.
- Added tests/assets/animated_skin.glb (3.4 KiB): six vertices, two joints, and a rotation + non-uniform-scale clip. The engine CPU test evaluates its midpoint; snapshot_animated_skin renders three exact keyframes through both skinned main/shadow pipelines. Local Vulkan SSIM is 1.0000 and GLES SSIM is 0.9833.

Run it with:

    cargo test --test gpu_examples snapshot_animated_skin -- --ignored --nocapture

Set BLADE_UPDATE_SNAPSHOTS=1 to intentionally regenerate the reference.

blade-engine/src/lib.rs
```diff
@@ -350,17 +364,18 @@ impl ops::IndexMut<JointHandle> for Physics {
 }
 
 struct Visual {
-    model: blade_asset::Handle<blade_render::Model>,
+    render_object: blade_render::Object,
```
kvark: I asked to not include the render object here - it has different meaning and there would be redundancy with this change.. Keep referring to the model as before.

blade-engine/src/lib.rs
```diff
@@ -350,17 +364,18 @@ impl ops::IndexMut<JointHandle> for Physics {
...
+    animation_model: Option<blade_asset::Handle<AnimationModel>>,
     similarity: nalgebra::geometry::Similarity3<f32>,
 }
 
 struct Object {
     name: String,
     rigid_body: rapier3d::dynamics::RigidBodyHandle,
-    prev_isometry: nalgebra::Isometry3<f32>,
```
kvark, line 360: we want to keep that

blade-render/code/raster.wgsl
```diff
@@ -26,6 +26,11 @@ struct RasterDrawParams {
     material: vec4<f32>,
 }
 
+struct SkinningParams {
+    // Affine transforms: four padded vec3 columns in uniform memory.
+    joint_matrices: array<mat4x3<f32>, 64>,
```
kvark: let's define this 64 constant somewhere explicitly

blade-render/code/raster.wgsl
```diff
@@ -53,6 +60,7 @@ struct VertexOutput {
 
 var<uniform> frame_params: RasterFrameParams;
 var<uniform> draw_params: RasterDrawParams;
+var<uniform> skinning_params: SkinningParams;
```
kvark: I wonder if this should be a storage buffer instead, and same for `shadow_skinning_params`. Benefits:
- no limit on the number of joints
- ability to take a pointer and pass it into `skin_matrix` method

blade-render/src/model/skinning.rs
```diff
@@ -0,0 +1,338 @@
...
+pub(crate) const MAX_JOINTS_PER_DRAW: usize = 64;
+const NO_INDEX: u32 = u32::MAX;
+
+/// Renderer-local bind transform and hierarchy link for one model node.
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct Node {
+    pub parent_index: u32,
+    pub translation: glam::Vec3,
+    pub rotation: glam::Quat,
+    pub scale: glam::Vec3,
```
kvark, line 17: do we really need the scale? aren't the skeletal animations just moving and rotating?

blade-render/src/model/skinning.rs
```diff
@@ -0,0 +1,338 @@
+use std::borrow::Cow;
+
+pub(crate) const MAX_JOINTS_PER_DRAW: usize = 64;
+const NO_INDEX: u32 = u32::MAX;
+
+/// Renderer-local bind transform and hierarchy link for one model node.
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct Node {
```
kvark: should this be called `BindNode` for clarity?

blade-render/src/raster/mod.rs
```diff
@@ -133,26 +165,40 @@ struct RasterSkyData {
...
 
+#[derive(Clone, Copy)]
+enum Variant {
+    Rigid,
+    Skinned,
+}
+
 struct RasterPipelines {
     main: gpu::RenderPipeline,
+    main_skinned: gpu::RenderPipeline,
```
kvark, line 214: do we need separate pipelines for this? Can't this be resolved in the shader itself, uniformly?

blade-render/src/render/mod.rs
```diff
@@ -1324,6 +1494,17 @@ impl RayTracer {
             geometry_count
         );
 
+        if !dynamic_blas_builds.is_empty() {
+            let mut encoder = command_encoder.acceleration_structure("animated BLAS");
+            for build in &dynamic_blas_builds {
+                encoder.build_bottom_level(
```
kvark: I wonder if it would help to recycle the BLAS and add a command encoder method to `update_bottom_level`? Re-creating or re-allocating constantly isn't the best.

blade-render/src/lib.rs
```diff
@@ -149,10 +174,23 @@ impl From<blade_asset::Handle<Model>> for Object {
...
     }
 }
 
+impl Object {
+    /// Set a fully evaluated pose for this instance.
+    ///
+    /// This only changes the current pose. The owner is responsible for
+    /// advancing [`Self::prev_pose`] together with [`Self::prev_transform`]
+    /// after rendering a frame.
+    pub fn set_pose(&mut self, pose: Pose) {
```
kvark: remove this

kvark: instead we can add a method to "flip": assign current transform and pose to the previous ones

blade-render/src/render/mod.rs
kvark: Implemented in 90c403b.

- `AccelerationStructureEncoder::update_bottom_level` (Vulkan `UPDATE` / Metal `refit`)
- `AccelerationStructureDesc::updatable` so the first build sets `ALLOW_UPDATE` / `Refit`
- Size query via `get_updatable_bottom_level_acceleration_structure_sizes` (`scratch` is `max(build, update)`)

The ray tracer keeps a 3-generation ring of per-instance vertex/transform/BLAS/scratch buffers. Current + previous stay alive for motion queries; the third is idle (previous frame has completed) and is refit in place. Topology mismatch reallocates and does a full build.

blade-render/src/model/skinning.rs
kvark: Kept scale. glTF rest poses and inverse-bind matrices are affine, and clips can animate scale (including non-uniform). The `animated_skin.glb` fixture is specifically a rotation + non-uniform-scale clip.

blade-render/src/lib.rs
kvark: `set_pose` is gone. `pose` stays public; `Object::flip()` copies current transform and pose into the previous-frame slots. Direct `blade-render` users call `flip()` after submit; the engine reconstructs submissions from `prev_isometry` / `visual.prev_pose` instead.

blade-engine/src/lib.rs
kvark: Kept `prev_isometry` on the engine object. Previous-frame transforms are `prev_isometry * visual.similarity` again, initialized to the spawn isometry rather than identity.

blade-render/code/raster.wgsl
kvark: Left as a uniform. WebGL2 has no vertex-stage storage buffers, and Naga rejects `ptr<uniform, …>` as a function argument (`InvalidArgumentPointerSpace`), so a shared `skin_matrix` pointer needs storage which the portable raster path cannot use.

The 64-joint cap stays the WebGL2 uniform-block limit. Uniform `mat4x3` is still 64 bytes per joint because `vec3` columns align to 16 bytes — packing Pose to 4x3 helps CPU clones, not this UBO.

blade-render/src/model/skinning.rs
kvark: Renamed to `BindNode`.

blade-render/src/raster/mod.rs
kvark: Kept the split. The earlier review wanted rigid draws not to execute skinning. Unifying still binds the 4 KiB identity palette and still fetches joint/weight attributes, so the extra pipelines are the cheap part of that split.

blade-engine/src/lib.rs
kvark: Done in 90c403b. `Visual` is back to a `model` handle (plus animation model, similarity, and `prev_pose`). The engine rebuilds a `render_objects` list each frame; the render object is a submission, not something stored on the visual.

blade-render/code/raster.wgsl
kvark: Named it `MAX_JOINTS_PER_DRAW` in `raster.wgsl`, matching the Rust constant.

blade-engine/src/animation.rs
```diff
@@ -0,0 +1,703 @@
...
+    let b = read_sample(channel, right, value_part);
+    let components = channel.property.component_count();
+    let mut result = [0.0; 4];
+    match channel.interpolation {
+        Interpolation::Step => result[..components].copy_from_slice(a),
+        Interpolation::Linear if channel.property == Property::Rotation => {
+            let qa = glam::Quat::from_array(a.try_into().unwrap()).normalize();
+            let mut qb = glam::Quat::from_array(b.try_into().unwrap()).normalize();
+            if qa.dot(qb) < 0.0 {
+                qb = -qb;
```
kvark: no need - glam's slerp will already do this for us

blade-engine/src/animation.rs
```diff
@@ -0,0 +1,703 @@
...
+        }
+    }
+    if channel.property == Property::Rotation {
+        result = glam::Quat::from_array(result).normalize().to_array();
+    }
+    result
+}
+
+fn evaluate(nodes: &[SkeletonNode], clip: &AnimationClip, time: f32) -> blade_render::Pose {
+    use AnimationProperty as Property;
```
kvark: no point in having this alias

blade-engine/src/lib.rs
```diff
@@ -537,25 +556,23 @@ impl Engine {
...
-                    },
-                    model: visual.model,
-                    color_tint: object.color_tint,
+            for visual in object.visuals.iter_mut() {
+                let current_pose = object.animation.and_then(|animation| {
+                    visual
+                        .animation_model
+                        .and_then(|handle| animation.pose(&animation_models[handle]))
                 });
+                let mut render_object = blade_render::Object::from(visual.model);
```
kvark: let's not even use `from` here. Just go straight into
```rust
render_object.push(blade_render::Object {
...
});
visual.prev_pose = current_pose;
```

blade-graphics/src/gles/mod.rs
```diff
@@ -120,7 +120,9 @@ pub struct Sampler {
 }
 
 #[derive(Clone, Copy, Debug, Default, Hash, PartialEq)]
-pub struct AccelerationStructure {}
+pub struct AccelerationStructure {
+    pub(super) updatable: bool,
```
kvark: Why would we need this? Should be the user contract: if they requested updateable, they can update. We don't need to remember it.

blade-graphics/src/metal/mod.rs
```diff
@@ -178,6 +178,7 @@ impl Sampler {
 #[derive(Clone, Copy, Debug, Hash, PartialEq)]
 pub struct AccelerationStructure {
     raw: *mut ProtocolObject<dyn metal::MTLAccelerationStructure>,
+    updatable: bool,
```
kvark, line 181: similarly, I really don't want this boolean to be there

blade-graphics/src/metal/resource.rs
```diff
@@ -88,7 +88,22 @@ impl super::Context {
         &self,
         meshes: &[crate::AccelerationStructureMesh],
     ) -> crate::AccelerationStructureSizes {
-        let descriptor = super::make_bottom_level_acceleration_structure_desc(meshes);
+        self.bottom_level_acceleration_structure_sizes(meshes, false)
+    }
+
+    pub fn get_updatable_bottom_level_acceleration_structure_sizes(
```
kvark: let's not have this method at all. The main public method should return both sizes for immutable and update-able, we don't want another method

blade-render/code/raster.wgsl
```diff
@@ -64,13 +77,48 @@ var shadow_tex: texture_depth_2d;
...
 var<uniform> shadow_draw_params: ShadowDrawParams;
+var<uniform> shadow_skinning_params: SkinningParams;
+
+fn unpack_joints(raw: u32) -> vec4<u32> {
+    return (vec4<u32>(raw) >> vec4<u32>(0u, 8u, 16u, 24u)) & vec4<u32>(0xFFu);
+}
+
+fn apply_affine(m: mat3x4<f32>, p: vec3<f32>) -> vec3<f32> {
+    let h = vec4<f32>(p, 1.0);
+    return vec3<f32>(dot(m[0], h), dot(m[1], h), dot(m[2], h));
```
kvark: why isn't this just a multiplication? `(h * m).xyz`

blade-render/src/model/mod.rs
```diff
@@ -1,12 +1,22 @@
...
 };
 
+#[cfg(feature = "asset")]
+use std::{
+    collections::hash_map::{Entry, HashMap},
+    hash,
+};
+
+mod skinning;
+pub(crate) use skinning::MAX_JOINTS_PER_DRAW;
```
kvark: might as well expose this to public

blade-render/src/render/mod.rs
```diff
@@ -1234,23 +1300,183 @@ impl RayTracer {
...
+                let vertex_slice = unsafe {
+                    std::slice::from_raw_parts_mut(
+                        vertex_buffer.data() as *mut crate::Vertex,
+                        vertex_count * 2,
+                    )
+                };
+                for geometry in &model.geometries {
+                    let range =
+                        geometry.vertex_range.start as usize..geometry.vertex_range.end as usize;
+                    model.skin_geometry_vertices(geometry, pose, &mut vertex_slice[range.clone()]);
```
kvark: why are we doing skinning on CPU at all?

blade-render/code/raster.wgsl
kvark: Yes — `h * m` for `mat3x4`. Replaced the three dots with that.

blade-engine/src/animation.rs
kvark: Removed. glam's `slerp` already negates when the dot product is negative.

blade-graphics/src/gles/mod.rs
kvark: Removed. The handle does not store `updatable`. `build_bottom_level` always uses `ALLOW_UPDATE` so a later `update_bottom_level` is a caller choice, not something we remember. Allocate `update_data` when you might refit.

blade-graphics/src/metal/mod.rs
kvark: Removed from the Metal handle as well.

blade-render/src/render/mod.rs
kvark: The BLAS has to be built from deformed triangles. Raster skinning lives in the vertex shader and never writes a buffer, so it cannot feed `update_bottom_level`.

The hit shader also cannot skin after a bind-pose BLAS: intersection would be against the rest pose.

A compute pass that writes the same vertex buffer the BLAS reads is the GPU version of this, and it only exists on the ray-tracing path (WebGL2 has no compute). I left CPU skinning as the simple RT path for now; a compute skin dispatch is the follow-up if we want this off the CPU.

blade-engine/src/animation.rs
kvark: Dropped the alias; `evaluate` matches on `AnimationProperty` directly.

blade-engine/src/lib.rs
kvark: Done — `render_objects.push(blade_render::Object { ... })` then `visual.prev_pose = current_pose`.

blade-render/src/model/mod.rs
kvark: `MAX_JOINTS_PER_DRAW` is now `pub` and re-exported from `blade_render::model`.

blade-graphics/src/metal/resource.rs
kvark: `get_updatable_bottom_level_acceleration_structure_sizes` is gone. `get_bottom_level_acceleration_structure_sizes` now returns both: `data`/`scratch` for a one-shot BLAS, `update_data`/`update_scratch` for a BLAS that may be refit (`update_scratch` covers the initial build and later updates).

blade-graphics/src/metal/resource.rs
```diff
@@ -88,7 +88,22 @@ impl super::Context {
         &self,
         meshes: &[crate::AccelerationStructureMesh],
     ) -> crate::AccelerationStructureSizes {
-        let descriptor = super::make_bottom_level_acceleration_structure_desc(meshes);
+        let immutable = self.bottom_level_acceleration_structure_sizes(meshes, false);
+        let updatable = self.bottom_level_acceleration_structure_sizes(meshes, true);
+        crate::AccelerationStructureSizes {
```
kvark: what if we keep this `Sizes` structure strictly as `data` and `scratch` like before, but the method would take a max from the immutable and the update-able? It's just numbers anyway, and a max would do just fine.

blade-graphics/src/metal/resource.rs
```diff
@@ -88,7 +88,22 @@ impl super::Context {
...
+            scratch: immutable.scratch,
+            update_data: updatable.data,
+            update_scratch: updatable.scratch.max(updatable.update_scratch),
+        }
+    }
+
+    fn bottom_level_acceleration_structure_sizes(
+        &self,
+        meshes: &[crate::AccelerationStructureMesh],
+        updatable: bool,
```
kvark: I think we should remove the argument and just make this method return the max sizes between immutable and updateable

kvark:
GPU skinning is in: a compute shader shares `skin.inc.wgsl` with the existing skinned vertex shaders.

- **Ray tracing:** required. Compute writes a device vertex buffer (current + previous pose), then the BLAS is built/refit from that. The CPU fill path is gone.
- **Vulkan/Metal raster:** optional. `Rasterizer::prepare()` skins once; shadow and main then draw with the rigid pipelines from that buffer. GLES keeps vertex-stage skinning (no compute).

blade-render/src/model/mod.rs
```diff
@@ -785,12 +931,12 @@ impl Baker {
             blade_graphics::AccelerationStructureDesc {
```
kvark, line 961: ok, I think we should add `updatable: bool` here like we used to.

kvark: `AccelerationStructureDesc::updatable` is back. Cooked model BLASes pass `false`; animated per-instance BLASes pass `true` so they can be refit.

blade-graphics/src/metal/resource.rs
kvark: Removed the `updatable` argument. The public size method queries both internally and returns the max.

blade-graphics/src/metal/resource.rs
kvark: `AccelerationStructureSizes` is back to `data` + `scratch`. The BLAS size query takes the max of a one-shot build and a refittable one, so callers allocate one pair either way.

blade-graphics/src/vulkan/resource.rs
```diff
@@ -261,21 +261,33 @@ impl super::Context {
...
-        let mut sizes_raw = vk::AccelerationStructureBuildSizesInfoKHR::default();
-        unsafe {
-            rt.acceleration_structure
-                .get_acceleration_structure_build_sizes(
-                    vk::AccelerationStructureBuildTypeKHR::DEVICE,
-                    &blas_input.build_info,
-                    &blas_input.max_primitive_counts,
-                    &mut sizes_raw,
-                )
+        let query = |updatable| {
```
kvark: could we organize this without a closure? e.g. in a loop?
```rust
let mut sizes = crate::AccelerationStructureSizes::default();
for updatable in [false,true] {
  ... 
  sizes.data = sizes.data.max(...);
  sizes.scratch = sizes.scratch.max(...);
}
```

Same goes for Metal implementation

blade-render/code/hit.inc.wgsl
```diff
@@ -22,10 +25,15 @@ struct IndexBuffer {
 struct HitEntry {
     index_buf: u32,
     vertex_buf: u32,
-    winding: f32,
+    prev_vertex_buf: u32,
     // packed quaternion
     geometry_to_world_rotation: u32,
+    winding: f32,
```
kvark: It's painful to waste 3 words on padding here.
Would it make sense to encode the winding somewhere? e.g. the sign of `geometry_to_world_rotation.w` or something?

blade-render/src/render/mod.rs
```diff
@@ -506,10 +509,19 @@ pub struct RayTracer {
...
+    /// Owned by the ray tracer until the next scene build.
     acceleration_structure: blade_graphics::AccelerationStructure,
+    /// Non-owning handle to the old scene used for this frame's motion queries.
+    /// Its ownership is transferred to the current `FrameResources` in
+    /// `build_scene` and it must never be destroyed directly by the ray tracer.
     prev_acceleration_structure: blade_graphics::AccelerationStructure,
+    /// Three generations of per-instance animated BLAS storage. The current
+    /// and previous generations stay alive for motion queries; the third is
+    /// idle and can be refit in place.
+    animated_blas_ring: AnimatedBlasRing,
```
kvark: why is this owned by the `RayTracer`? I'd expect BLAS to be owned by `blade_render::Object`

blade-graphics/src/vulkan/resource.rs
kvark: Done — both Vulkan and Metal now loop `[false, true]` and `max` into a default `AccelerationStructureSizes`.

blade-render/src/render/mod.rs
kvark: Moved. Each `Object` now owns an `ObjectBlas` with the three-generation vertex/BLAS ring. `build_scene` takes `&mut Object` and refits in place. The engine parks that storage on the visual between frames; callers must `ObjectBlas::destroy` before dropping the object.

blade-render/code/hit.inc.wgsl
kvark: Encoded winding in the sign of the packed quaternion's `w` (`q` and `-q` are the same rotation). That drops the `winding` field and the three padding words, so `HitEntry` is 256 bytes.

blade-graphics/src/vulkan/command.rs
```diff
@@ -942,6 +942,43 @@ impl crate::traits::AccelerationStructureEncoder
...
+    ) {
+        self.encode_bottom_level(
+            acceleration_structure,
+            meshes,
+            scratch_data,
+            vk::BuildAccelerationStructureModeKHR::BUILD,
+        );
+    }
+
+    fn update_bottom_level(
```
kvark, line 969: do we even need the `update` given the same signature? sounds like we already know if we can update in place based on the validity of `acceleration_structure` and it having an `updatable` flag

blade-render/code/hit.inc.wgsl
```diff
@@ -10,6 +10,9 @@ struct Vertex {
     tex_coords: vec2<f32>,
     normal: u32,
     tangent: u32,
+    joints: u32,
+    weights: u32,
+    _pad: vec2<u32>,
```
kvark: need a TODO to optimize this. With this padding and a bitangent_sign we are close of saving a whole vec4 here.
Or maybe you can suggest some ways right here?

blade-render/code/hit.inc.wgsl
```diff
@@ -50,6 +54,16 @@ fn decode_normal(raw: u32) -> vec3<f32> {
     return unpack4x8snorm(raw).xyz;
 }
 
+fn hit_geometry_rotation(entry: HitEntry) -> vec4<f32> {
+    let q = unpack4x8snorm(entry.geometry_to_world_rotation);
+    return normalize(vec4<f32>(q.xyz, abs(q.w)));
```
kvark: This is concerning. I don't think we should provide an *invalid* quaternion.
The client should be able to use it as is without any special treatment. But additionally, one can extract the sign from `q.w` to know the handed ness.

blade-render/src/model/mod.rs
```diff
@@ -121,6 +153,10 @@ pub struct Model {
     pub index_buffer: blade_graphics::Buffer,
     pub transform_buffer: blade_graphics::Buffer,
     pub acceleration_structure: blade_graphics::AccelerationStructure,
+    pub(crate) nodes: Vec<BindNode>,
+    pub(crate) bind_pose: Pose,
+    pub(crate) skins: Vec<Skin>,
+    pub(crate) vertices: Vec<crate::Vertex>,
```
kvark: We should only be storing them when on WebGL. Maybe leave a comment that this is a compatibility mode thing unused on the main path?

blade-render/src/raster/mod.rs
```diff
@@ -436,16 +536,96 @@ impl Rasterizer {
...
+        for (object_index, object) in objects.iter().enumerate() {
+            let model = &asset_hub.models[object.model];
+            if !skin::model_needs_vertex_skin(model) {
+                continue;
+            }
+            let vertex_count = model.vertices.len();
+            let vertex_size = (vertex_count * mem::size_of::<Vertex>()) as u64;
+            let reuse = slots
+                .get(dynamic_index)
+                .is_some_and(|slot| slot.matches(vertex_count));
```
kvark: when would it not match? shouldn't this be guaranteed and expected?

blade-render/src/raster/mod.rs
```diff
@@ -504,25 +693,33 @@ impl Rasterizer {
     }
 
     #[profiling::function]
-    pub fn render(
+    pub fn render<'a>(
         &mut self,
         pass: &mut gpu::RenderCommandEncoder,
         camera: &crate::Camera,
-        objects: &[Object],
+        objects: impl IntoIterator<Item = &'a Object>,
```
kvark: what's wrong with `&[Object]?`

blade-graphics/src/vulkan/command.rs
kvark: We still need both encoder ops. The first fill of an updatable BLAS has to be a `BUILD` with `ALLOW_UPDATE`; a later `UPDATE`/`refit` is only valid after that. The handle is already non-null on that first build (we create it, then encode), so validity plus `updatable` is not enough to choose the mode. The renderer keeps the reuse bit for that, and `update_bottom_level` stays the explicit refit.

blade-render/src/model/mod.rs
kvark: Dropped the CPU copy on every backend, including WebGL. After cooking, bind-pose vertices live only in `vertex_buffer`; `Model::vertex_count()` is derived from geometry ranges. GLES still skins in the vertex shader from that GPU buffer.

blade-render/src/raster/mod.rs
kvark: That was the old renderer-wide ring keyed by draw order, so a different object (or a reloaded model) in the same slot could miss. Dest buffers now live on `Object::skin`, same idea as `Object::blas`. The only realloc is when that object's vertex count changes.

blade-render/code/hit.inc.wgsl
kvark: `q` and `-q` are the same rotation, so the packed value is a valid quaternion as-is. `hit_geometry_rotation` now only unpacks and normalizes — no `abs(w)`. Winding is `sign(w)` (zero counts as +1 so a degenerate pack does not zero the geometric normal).

blade-render/src/raster/mod.rs
kvark: Nothing — `render` and `render_directional_shadows` now take `&[Object]`. `prepare` / `build_scene` take `&mut [Object]` so they can fill the per-object GPU storage.

blade-render/code/hit.inc.wgsl
kvark: Left a TODO on `Vertex` in `hit.inc.wgsl` / `skin.inc.wgsl`.

The slack is `bitangent_sign` plus the 8-byte WGSL pad. Two follow-ups that actually get us a vec4 back:
- put `bitangent_sign` in the unused snorm `w` of packed `normal`
- drop `joints`/`weights` from the *post-skin* dest buffer (the compute pass already applied them), which with the above is a 32-byte RT vertex

blade-engine/src/lib.rs
```diff
@@ -537,30 +558,46 @@ impl Engine {
...
+                    blas: mem::take(&mut visual.blas),
+                    skin: mem::take(&mut visual.skin),
                 });
+                visual.prev_pose = current_pose;
             }
             object.prev_isometry = isometry;
         }
     }
 
+    fn restore_object_blas(
```
kvark: this looks wrong to me. Data should be flowing from the engine into the renderer strictly one way, not going back

blade-engine/src/lib.rs
```diff
@@ -537,30 +558,46 @@ impl Engine {
...
+                    transform: make_render_transform(
+                        (isometry * visual.similarity).to_homogeneous(),
+                    ),
+                    prev_transform: make_render_transform(
+                        (object.prev_isometry * visual.similarity).to_homogeneous(),
+                    ),
                     color_tint: object.color_tint,
+                    pose: current_pose.clone(),
+                    prev_pose: visual.prev_pose.take(),
+                    blas: mem::take(&mut visual.blas),
```
kvark: this should just be a plain copy or a reference, not extraction

blade-render/code/hit.inc.wgsl
```diff
@@ -50,6 +57,15 @@ fn decode_normal(raw: u32) -> vec3<f32> {
     return unpack4x8snorm(raw).xyz;
 }
 
+fn hit_geometry_rotation(entry: HitEntry) -> vec4<f32> {
+    return normalize(unpack4x8snorm(entry.geometry_to_world_rotation));
+}
+
+fn hit_winding(entry: HitEntry) -> f32 {
+    let w = unpack4x8snorm(entry.geometry_to_world_rotation).w;
```
kvark: surely there is a more efficient way here? something like `select(-1.0, 1.0, entry.geometry_to_world_rotation&0xFFu >= 0x80u)` depending on where `w` is based on the layout

blade-render/src/model/mod.rs
```diff
@@ -121,6 +153,18 @@ pub struct Model {
     pub index_buffer: blade_graphics::Buffer,
     pub transform_buffer: blade_graphics::Buffer,
     pub acceleration_structure: blade_graphics::AccelerationStructure,
+    pub(crate) nodes: Vec<BindNode>,
```
kvark: ok, this isn't really a `BindNode`. The `Bind` word is overloaded. Maybe a `SkinNode`?

blade-render/src/raster/mod.rs
```diff
@@ -249,16 +285,60 @@ impl RasterPipelines {
...
+        gpu.destroy_buffer(self.vertex_buffer);
+    }
+}
+
+/// Per-object GPU storage for compute-skinned raster vertices.
+///
+/// Persist this with the [`crate::Object`] across frames. Call
+/// [`Self::destroy`] before dropping it.
+#[derive(Default)]
+pub struct ObjectSkin {
```
kvark, line 421: should we have an abstraction for this? e.g. `RollingBuffer<T, N>` or something?

kvark: Added `RollingBuffer<T, N>` in `blade-render` util. `ObjectBlas` is a 3-slot ring and `ObjectSkin` is a 2-slot ring on top of it.

blade-engine/src/lib.rs
kvark: Removed the restore path. Visuals no longer hold GPU instance state. Each visual has a stable `instance_id`; the engine keeps `render_objects` and only writes CPU fields (transform, pose, …) onto the matching `Object`. Leftover instances are destroyed, nothing flows back from the renderer.

blade-engine/src/lib.rs
kvark: Dropped the `mem::take`. GPU storage stays on the persisted `Object`. The engine looks up that object by instance id and overwrites the CPU fields; `blas`/`skin` are left alone.

blade-render/code/hit.inc.wgsl
kvark: `w` is the high byte of the little-endian packed snorm quat, so this is now `select(-1.0, 1.0, (entry.geometry_to_world_rotation >> 24u) < 0x80u)` — no unpack.

blade-render/src/model/mod.rs
kvark: Renamed to `SkinNode`. `bind_pose` stays as the evaluated rest pose.

blade-render/code/hit.inc.wgsl
kvark: Even simpler: `select(-1.0, 1.0, packed < 0x80000000u)` — the snorm sign of `w` is bit 31 of the word.

blade-engine/src/animation.rs
```diff
@@ -0,0 +1,701 @@
...
+    // `state` marks nodes as unvisited, on the current chain, or resolved.
+    let mut globals = vec![glam::Mat4::IDENTITY; nodes.len()];
+    let mut state = vec![0u8; nodes.len()];
+    for start in 0..nodes.len() {
+        let mut chain = Vec::new();
+        let mut current = start;
+        let mut is_cycle = false;
+        loop {
+            if state[current] != 0 {
+                is_cycle = state[current] == 1;
```
kvark: why not just assert right here and spare `is_cycle` variable entirely?

blade-engine/src/animation.rs
```diff
@@ -0,0 +1,701 @@
...
+            AnimationProperty::Translation => node.translation = glam::Vec3::from_slice(&value),
+            AnimationProperty::Rotation => node.rotation = glam::Quat::from_array(value),
+            AnimationProperty::Scale => node.scale = glam::Vec3::from_slice(&value),
+        }
+    }
+
+    // Resolve the global transforms with constant-time cycle detection:
+    // `state` marks nodes as unvisited, on the current chain, or resolved.
+    let mut globals = vec![glam::Mat4::IDENTITY; nodes.len()];
+    let mut state = vec![0u8; nodes.len()];
```
kvark: would make sense to declare a local enum type for this state

blade-engine/src/lib.rs
```diff
@@ -524,41 +552,90 @@ impl Engine {
...
-        physics: &mut Physics,
-        objects: &mut slab::Slab<Object>,
-        render_objects: &mut Vec<blade_render::Object>,
+        physics: &Physics,
+        objects: &slab::Slab<Object>,
+        render_objects: &mut [blade_render::Object],
+        retired_render_objects: &mut Vec<blade_render::Object>,
+        animation_models: &blade_asset::AssetManager<animation::Baker>,
         time_ahead: f32,
+        temp: &mut blade_render::FrameResources,
```
kvark: this and the `retired_render_objects` seem unrelated. We need to either move it out into another method, or just inline manually, since it's pretty trivial

blade-render/code/hit.inc.wgsl
```diff
@@ -4,12 +4,18 @@
 // and `debug-param.inc.wgsl` for the texture flags.
 
 // Has to match the host!
+// TODO: 48 bytes is close to a spare vec4. `bitangent_sign` could live in the
```
kvark: Let's implement this. Perhaps we need to treat the joints+weights as a separate vertex buffer from the start, so that the difference between skinned/unskinned is simply this extra binding, and we can share the same vertex struct for the base?

blade-render/code/raster.wgsl
```diff
@@ -71,34 +64,64 @@ fn raster_shadow_vs(input: Vertex) -> @builtin(position) vec4<f32> {
...
     out.clip_pos = frame_params.view_proj * pos_world;
     out.world_pos = pos_world.xyz;
-    let n = normalize(quat_rotate(draw_params.normal_quat, decode_normal(input.normal)));
-    let t = normalize(quat_rotate(draw_params.normal_quat, decode_normal(input.tangent)));
-    let b = normalize(cross(n, t)) * input.bitangent_sign;
+    let linear = mat3x3<f32>(draw_params.model[0].xyz, draw_params.model[1].xyz, draw_params.model[2].xyz);
+    let n = normalize(draw_params.normal_matrix * normal);
+    let transformed_tangent = linear * tangent;
+    let t = normalize(transformed_tangent - n * dot(n, transformed_tangent));
+    let b = normalize(cross(n, t)) * bitangent_sign * sign(determinant(linear));
```
kvark: this code is also seen in `hit_tangent_space`, can we put it into a shared header instead?

blade-render/code/skin.inc.wgsl
```diff
@@ -0,0 +1,60 @@
...
+// Must match `blade_render::model::MAX_JOINTS_PER_DRAW` and `blade_render::Vertex`.
+const MAX_JOINTS_PER_DRAW: u32 = 64;
+
+struct SkinningParams {
+    post_transform: mat3x4<f32>,
+    joint_matrices: array<mat3x4<f32>, MAX_JOINTS_PER_DRAW>,
+}
+
+// TODO: see Vertex in hit.inc.wgsl — bitangent_sign and the 8-byte pad are
+// the remaining slack toward a 32-byte vertex.
```
kvark: we are using a different vertex definition here? can't we share it?

blade-render/src/raster/mod.rs
```diff
@@ -89,7 +117,7 @@ struct RasterFrameParams {
 #[derive(Clone, Copy, bytemuck::Zeroable, bytemuck::Pod)]
 struct RasterDrawParams {
     model: [f32; 16],
-    normal_quat: [f32; 4],
+    normal_matrix: [f32; 12],
```
kvark: this looks weird. Why would we want anything but a quaternion for a rotation here?

blade-render/src/raster/mod.rs
```diff
@@ -465,28 +694,51 @@ impl Rasterizer {
...
+                    };
+                    if skinned {
+                        pc.bind(
+                            0,
+                            &RasterShadowSkinnedData {
+                                shadow_frame_params: ShadowFrameParams {
+                                    light_view_proj: light_view_proj.to_cols_array(),
+                                },
+                                shadow_draw_params,
+                                skinning_params: skin::make_skinning_params(model, geometry, pose),
```
kvark, line 783: should skinning params just bind separately, e.g. to bind slot 1?

docs/CHANGELOG.md
```diff
@@ -2,6 +2,25 @@ Changelog for *Blade* project
 
 ## (TBD)
 
+- render/engine: load glTF skins and TRS animation clips, including step,
```
kvark: can we write CHANGELOG in a more concise way? too much prose here

blade-render/src/raster/mod.rs
```diff
@@ -546,49 +801,64 @@ impl Rasterizer {
...
+                        material: [normal_scale, material.metalness, material.roughness, 0.0],
+                    };
+                    let base_color_tex = texture_or_white(material.base_color_texture);
+                    let metallic_roughness_tex =
+                        texture_or_white(material.metallic_roughness_texture);
+                    let emissive_tex = texture_or_white(material.emissive_texture);
+                    if skinned {
+                        pc.bind(
+                            0,
+                            &RasterMainSkinnedData {
```
kvark: can't we also split this into the general data and the skinned extension that is bound to slot 1 ?

blade-render/src/render/mod.rs
```diff
@@ -1157,15 +1256,21 @@ impl RayTracer {
     }
 
     #[profiling::function]
-    pub fn build_scene(
+    pub fn build_scene<'a>(
         &mut self,
         command_encoder: &mut blade_graphics::CommandEncoder,
-        objects: &[crate::Object],
+        objects: &'a mut [crate::Object],
```
kvark: what's the point of this lifetime being explicit?

PR #377, opened 2026-08-27
Add raster point lights and window-path particles

kvark:
Forward shading picks one of up to eight local lights (best score, with a stochastic alternative). Particle systems now simulate and draw in the desktop raster pass, not only XR.

### DIFF
--- a/blade-engine/src/lib.rs
+++ b/blade-engine/src/lib.rs
@@ -494,7 +494,8 @@ pub struct Engine {
     choir: Arc<choir::Choir>,
     data_path: String,
     time_ahead: f32,
-    last_dt: f32,
+    particle_clock: f32,
+    last_particle_clock: f32,
 }
 
 impl Engine {
@@ -785,7 +786,8 @@ impl Engine {
             choir,
             data_path: config.data_path.clone(),
             time_ahead: 0.0,
-            last_dt: 0.016,
+            particle_clock: 0.0,
+            last_particle_clock: 0.0,
         }
     }
 
@@ -819,7 +821,7 @@ impl Engine {
     #[profiling::function]
     pub fn update(&mut self, dt: f32) {
         self.choir.check_panic();
-        self.last_dt = dt;
+        self.particle_clock += dt;
         self.time_ahead += dt;
         while self.time_ahead >= self.physics.integration_params.dt {
             self.physics.step();
@@ -878,6 +880,12 @@ impl Engine {
         }
     }
 
+    fn take_particle_dt(&mut self) -> f32 {
+        let dt = (self.particle_clock - self.last_particle_clock).clamp(0.0, 0.05);
+        self.last_particle_clock = self.particle_clock;
+        dt
+    }
+
     /// Drain contact events from the last physics update.
     pub fn drain_contacts(&mut self) -> impl Iterator<Item = ContactEvent> + '_ {
         self.contact_events.drain(..)
@@ -913,7 +921,8 @@ impl Engine {
             ..
         } = self.renderer
         {
-            raster_config.point_lights = lights.to_vec();
+            raster_config.point_lights.clear();
+            raster_config.point_lights.extend_from_slice(lights);
         }
     }
 
@@ -927,6 +936,7 @@ impl Engine {
         physical_size: winit::dpi::PhysicalSize<u32>,
         scale_factor: f32,
     ) {
+        let particle_dt = self.take_particle_dt();
         self.hot_reload_if_needed();
 
         // Note: the resize is split in 2 parts because `wait_for_previous_frame`
@@ -1023,10 +1033,11 @@ impl Engine {
             }
         }
 
-        if let Some(ref pipeline) = self.particle_pipeline {
-            let dt = self.last_dt.clamp(0.001, 0.05);
+        if particle_dt > 0.0
+            && let Some(ref pipeline) = self.particle_pipeline
+        {
             for (_, system) in self.particle_systems.iter_mut() {
-                system.update(pipeline, command_encoder, dt);
+                system.update(pipeline, command_encoder, particle_dt);
             }
         }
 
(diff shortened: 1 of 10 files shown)


blade-render/code/raster.wgsl
```diff
@@ -129,6 +136,73 @@ fn directional_shadow(world_pos: vec3<f32>, n: vec3<f32>) -> f32 {
...
+    var sum_score = 0.0;
+    for (var i = 0u; i < 8u; i++) {
+        if (i >= count) {
+            scores[i] = 0.0;
+            continue;
+        }
+        let score = point_light_score(frame_params.point_lights[i], world_pos, n);
+        scores[i] = score;
+        sum_score += score;
+        if (score > best_score) {
```
kvark: we should probably treat this as a reservoir that picks a random light from the list, instead of strictly picking the best light at all times.

blade-render/code/raster.wgsl
```diff
@@ -129,6 +136,73 @@ fn directional_shadow(world_pos: vec3<f32>, n: vec3<f32>) -> f32 {
...
+    let count = u32(frame_params.settings.z);
+    if (count == 0u) {
+        return vec3<f32>(0.0);
+    }
+
+    var scores: array<f32, 8>;
+    var best_i = 0u;
+    var best_score = 0.0;
+    var sum_score = 0.0;
+    for (var i = 0u; i < 8u; i++) {
```
kvark: can we have this as a constant somewhere?

blade-render/code/raster.wgsl
```diff
@@ -129,6 +136,73 @@ fn directional_shadow(world_pos: vec3<f32>, n: vec3<f32>) -> f32 {
...
+        if (i >= count) {
+            scores[i] = 0.0;
+            continue;
+        }
+        let score = point_light_score(frame_params.point_lights[i], world_pos, n);
```
kvark: we need some sort of an acceleration structure to look up which lights are interesting and potentially viable. I guess we can follow-up on that once we have more lights in a scene, but let's leave a todo for this anyway now

blade-render/src/raster/mod.rs
```diff
@@ -17,6 +47,10 @@ pub struct RasterConfig {
     pub space_sky: bool,
     /// Optional real-time directional shadow-map effect.
     pub directional_shadows: Option<DirectionalShadowConfig>,
+    /// Local lights considered by the forward pass. Only the first
+    /// `point_light_count` entries are used.
+    pub point_lights: [PointLight; MAX_POINT_LIGHTS],
```
kvark: should probably be just a `Vec`? the actual limitation can be applied down the stack

blade-render/src/raster/mod.rs
```diff
@@ -780,19 +834,44 @@ impl Rasterizer {
                 env_map_enabled as u32 as f32,
                 // the surface may expect us to encode the values ourselves
                 (self.color_space == gpu::ColorSpace::Srgb) as u32 as f32,
-                0.0,
-                0.0,
+                config.point_light_count.min(MAX_POINT_LIGHTS as u32) as f32,
+                pos.x + pos.y * 1.37 + pos.z * 9.17,
```
kvark: this magic equation deserves its own function with meaning

blade-render/code/raster.wgsl
```diff
@@ -12,9 +17,11 @@ struct RasterFrameParams {
     // w component is a flag for the procedural space sky
     ambient_color: vec4<f32>,
     // x: environment map enabled, y: the surface needs sRGB encoding
+    // z: point light count, w: stochastic seed
     settings: vec4<f32>,
     // x: enabled, y: strength, z: receiver normal bias, w: texel size
     shadow_params: vec4<f32>,
+    point_lights: array<PointLight, 8>,
```
kvark: needs to be a proper constant

kvark: Would also make sense to add a variant like `Expansion::Size(u32)` and register this constant

kvark:
Addressed the review in 1678c32:

- Local lights live in their own UBO so `sky_params` stays 288 bytes (this was the `snapshot_space_sky` panic).
- `RasterConfig::point_lights` is a `Vec`; the `MAX_POINT_LIGHTS` cap is applied when packing.
- Fragment shading uses a weighted reservoir instead of always taking the best light.
- `MAX_POINT_LIGHTS` is a registered `Expansion::Size`.
- Added a TODO for a spatial structure once scenes have more lights.
- Seed mix is `stochastic_light_seed`.
- Android asteroids now uses `..Default::default()`.

blade-engine/src/lib.rs
```diff
@@ -900,6 +903,20 @@ impl Engine {
...
+    ///
+    /// The fragment shader samples one of up to `MAX_POINT_LIGHTS` of these
+    /// with probability proportional to a local score.
+    pub fn set_point_lights(&mut self, lights: &[blade_render::PointLight]) {
+        if let Renderer::Rasterizer {
+            ref mut raster_config,
+            ..
+        } = self.renderer
+        {
+            raster_config.point_lights = lights.to_vec();
```
kvark: don't re-allocate the vector, clear and extend instead

blade-engine/src/lib.rs
```diff
@@ -1006,6 +1023,13 @@ impl Engine {
             }
         }
 
+        if let Some(ref pipeline) = self.particle_pipeline {
+            let dt = self.last_dt.clamp(0.001, 0.05);
```
kvark: this logic looks wrong to me, it doesn't respect `update()` called multiple times. We should just remember the last particle update time here and compute the delta

blade-engine/src/lib.rs
```diff
@@ -1113,7 +1137,7 @@ impl Engine {
                         &render_camera,
                         &self.render_objects,
                         &self.asset_hub,
-                        *raster_config,
```
kvark, line 1116: can't we borrow it normally? why clone?

kvark:
Follow-up in 22ad679:

- GPU CI: `snapshot_space_sky` now expands `#use` via the snapshot helper (that was `needs host expansions`).
- `set_point_lights` clears/extends the existing `Vec`.
- Particle dt is time since the last render, accumulated across `update()` calls.
- `Rasterizer::{render, render_directional_shadows, render_sky_only}` take `&RasterConfig`.

PR #378, opened 2026-08-28
Fix WebGL2 buffer classes, empty texture uploads, and 0x0 storage

kvark:
The Redline web console was full of real WebGL errors, not just noise:

- `texSubImage: Desired upload requires more bytes than are available (0)`
- `Index buffer too small`
- `Buffer previously bound to non-ELEMENT_ARRAY_BUFFER cannot be now bound to ELEMENT_ARRAY_BUFFER`
- `texStorage: Dimensions must be non-zero`
- comparison sampling with LINEAR

Cause: the egui `BufferBelt` mixed vertex, index, and texture staging in one WebGL buffer. WebGL assigns a bind class on first bind, and the belt never `sync_buffer`d after CPU writes (native GLES uses persistent maps). Later draws and PIXEL_UNPACK copies then saw the wrong target or a 0-byte buffer.

This splits data vs index belts, syncs after CPU writes, rejects 0×0 texStorage, resets `UNPACK_ROW_LENGTH`, and uses NEAREST for comparison samplers.

`Particle compute is not available` remains: WebGL2 has no compute. That path already no-ops.

### DIFF
--- a/blade-egui/src/lib.rs
+++ b/blade-egui/src/lib.rs
@@ -329,8 +329,8 @@ impl GuiPainter {
         }
 
         self.triage_deletions(context);
-        self.data_belt.trim(32, context);
-        self.index_belt.trim(32, context);
+        self.data_belt.trim(4, context);
+        self.index_belt.trim(4, context);
     }
 
     /// Render the set of clipped primitives into a render pass.
--- a/blade-egui/src/lib.rs
+++ b/blade-egui/src/lib.rs
@@ -407,8 +407,6 @@ impl GuiPainter {
                 );
             }
         }
-        self.data_belt.sync(context);
-        self.index_belt.sync(context);
     }
 
     /// Call this after submitting work at the given `sync_point`.
--- a/blade-egui/src/lib.rs
+++ b/blade-egui/src/lib.rs
@@ -329,8 +329,8 @@ impl GuiPainter {
         }
 
         self.triage_deletions(context);
-        self.data_belt.trim(4, context);
-        self.index_belt.trim(4, context);
+        self.data_belt.trim(32, context);
+        self.index_belt.trim(32, context);
     }
 
     /// Render the set of clipped primitives into a render pass.
--- a/blade-egui/src/lib.rs
+++ b/blade-egui/src/lib.rs
@@ -144,8 +144,9 @@ impl GuiTexture {
 /// It can render egui primitives into a render pass.
 pub struct GuiPainter {
     pipeline: blade_graphics::RenderPipeline,
-    //TODO: find a better way to allocate temporary buffers.
-    belt: BufferBelt,
+    /// Vertices and texture staging. WebGL forbids mixing this with index data.
+    data_belt: BufferBelt,
+    index_belt: BufferBelt,
     textures: HashMap<egui::TextureId, GuiTexture>,
     //TODO: this could also look better
     textures_dropped: Vec<GuiTexture>,
@@ -156,7 +157,8 @@ impl GuiPainter {
     /// Destroy the contents of the painter.
     pub fn destroy(&mut self, context: &blade_graphics::Context) {
         context.destroy_render_pipeline(&mut self.pipeline);
-        self.belt.destroy(context);
+        self.data_belt.destroy(context);
+        self.index_belt.destroy(context);
         for (_, gui_texture) in self.textures.drain() {
             gui_texture.delete(context);
         }
@@ -214,15 +216,23 @@ impl GuiPainter {
             multisample_state: Default::default(),
         });
 
-        let belt = BufferBelt::new(BufferBeltDescriptor {
+        let data_belt = BufferBelt::new(BufferBeltDescriptor {
             memory: blade_graphics::Memory::Shared,
             min_chunk_size: 0x1000,
-            alignment: blade_graphics::limits::STORAGE_BUFFER_ALIGNMENT,
+            alignment: 4,
+            target: blade_graphics::BufferTarget::Data,
+        });
+        let index_belt = BufferBelt::new(BufferBeltDescriptor {
+            memory: blade_graphics::Memory::Shared,
+            min_chunk_size: 0x1000,
+            alignment: 4,
+            target: blade_graphics::BufferTarget::Index,
         });
 
         Self {
             pipeline,
-            belt,
+            data_belt,
+            index_belt,
             textures: Default::default(),
             textures_dropped: Vec::new(),
             textures_to_delete: Vec::new(),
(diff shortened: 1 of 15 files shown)


blade-graphics/src/gles/resource.rs
```diff
@@ -339,7 +345,13 @@ impl crate::traits::ResourceDevice for super::Context {
             glow::TEXTURE_WRAP_T,
             glow::TEXTURE_WRAP_R,
         ];
-        let (min, mag) = map_filter_modes(desc.min_filter, desc.mag_filter, desc.mipmap_filter);
+        // WebGL2 comparison sampling with LINEAR is implementation-defined
```
kvark: sounds fine to me? we shouldn't switch to NEAREST just because LINEAR isn't always available

blade-util/src/belt.rs
```diff
@@ -88,6 +95,9 @@ impl BufferBelt {
         unsafe {
             std::ptr::copy_nonoverlapping(data.as_ptr(), bp.data(), data.len());
         }
+        // WebGL has no persistent mapping. Re-upload the chunk so the GPU
+        // sees the CPU writes before the next draw or texture copy.
+        gpu.sync_buffer(bp.buffer, self.desc.target);
```
kvark: is this going to re-sync the whole buffer on every upload? perhaps we should do it once before rendering the frames instead?

blade-util/src/belt.rs
```diff
@@ -12,6 +12,13 @@ pub struct BufferBeltDescriptor {
     pub memory: gpu::Memory,
     pub min_chunk_size: u64,
     pub alignment: u64,
+    /// WebGL2 bind class for every chunk in this belt.
```
kvark: no need for these comments, people can look it up in the type itself

blade-egui/src/lib.rs
```diff
@@ -214,15 +216,23 @@ impl GuiPainter {
...
+        let data_belt = BufferBelt::new(BufferBeltDescriptor {
             memory: blade_graphics::Memory::Shared,
             min_chunk_size: 0x1000,
             alignment: blade_graphics::limits::STORAGE_BUFFER_ALIGNMENT,
+            target: blade_graphics::BufferTarget::Data,
+        });
+        let index_belt = BufferBelt::new(BufferBeltDescriptor {
+            memory: blade_graphics::Memory::Shared,
+            min_chunk_size: 0x1000,
+            alignment: blade_graphics::limits::STORAGE_BUFFER_ALIGNMENT,
```
kvark: we are no longer binding it as storage, are we?

kvark: Right — these are vertex/index (and texture staging) chunks, not storage buffers. Alignment is 4 now.

blade-util/src/belt.rs
kvark: Removed; `BufferTarget` already documents the bind classes.

blade-graphics/src/gles/resource.rs
kvark: Agreed — reverted. LINEAR comparison sampling stays as requested.

blade-util/src/belt.rs
kvark: Now `BufferBelt::sync` uploads all active chunks once, called at the end of `GuiPainter::paint` after every alloc for the frame.

blade-egui/src/lib.rs
```diff
@@ -394,6 +407,8 @@ impl GuiPainter {
                 );
             }
         }
+        self.data_belt.sync(context);
```
kvark: this is too late, isn't it?
I wonder if there even a path here that would work best without changing the API. I could think of a way to pass a range into `sync_buffer`, but perhaps it's not the only one.

kvark: You're right — `update_textures` records copies before `paint`, so a trailing sync is too late.

Went with a range on `sync_buffer`: `sync_buffer_range` uploads just the new piece at alloc time (allocate the chunk on first WebGL bind, then `bufferSubData`). No extra painter API; draws and copies see GPU data immediately.

blade-graphics/src/gles/resource.rs
```diff
@@ -121,25 +121,32 @@ impl crate::traits::ResourceDevice for super::Context {
...
+            crate::BufferTarget::Data => glow::ARRAY_BUFFER,
+            crate::BufferTarget::Index => glow::ELEMENT_ARRAY_BUFFER,
+        };
+        unsafe {
+            let data = slice::from_raw_parts(piece.data(), size as usize);
+            gl.bind_buffer(raw_target, piece.buffer.raw);
+            // WebGL starts with no storage. Allocate the full chunk on first
+            // bind so later ranges can `bufferSubData` without orphaning.
+            #[cfg(target_arch = "wasm32")]
+            if gl.get_buffer_parameter_i32(raw_target, glow::BUFFER_SIZE) == 0 {
```
kvark: unfortunate check, but I understand why we are doing it and can't think of anything better
we can probably also rule that out by checking the offset == 0 to avoid calling this GL function too much

kvark (commented):
This technically works, but the cost is one more API function.
I rubs me a bit that all of that machinery isn't ideal for GLES either: if you are uploading data every time before you are using it, you might as well just upload it to the start of the buffer, since the upload is effectively a part of the command stream.
At least, if we want to prevent the GL driver from renaming that buffer internally, there must be a specific alignment that we respect when uploading those subranges?
Let's check the MESA sources to verify.

kvark:
Checked Mesa ([buffer mapping patterns](https://docs.mesa3d.org/gallium/buffermapping.html) and `u_upload_mgr`).

Valid range is a prefix from 0. Incrementing `bufferSubData` (Portal 2) works if the driver tracks that prefix or blits via `u_upload_mgr` (VBO/IBO suballoc alignment is typically 4, `PIPE_CAP_VERTEX_BUFFER_OFFSET`). `glBufferData` is the explicit orphan/rename path (Terraria, Don’t Starve).

Since the GLES encoder is deferred and we upload before use, packing subranges isn’t a win and wasn’t worth another API. `512f1ae` drops `sync_buffer_range`: each belt alloc is a dedicated buffer at offset 0, and `sync_buffer` is `glBufferData` from the start.

blade-graphics/src/gles/resource.rs
kvark: Dropped the query. Each belt alloc is now a dedicated buffer at offset 0, and GLES `sync_buffer` does `glBufferData` from the start — first bind allocates and locks the WebGL class, later calls orphan.

kvark:
Switched off the per-alloc orphan path. Packed 4-byte-aligned `bufferSubData` is the one that matches Vulkan/Metal and Mesa (`u_upload_mgr` / `PIPE_CAP_VERTEX_BUFFER_OFFSET`).

`c5f3f67`: belt packs again, `sync_buffer` takes a piece range (no extra function), GLES allocates storage only when `offset == 0` then `bufferSubData`s the aligned subrange.

PR #379, opened 2026-08-29
Fix WebGL color space

kvark:
WebGL's canvas drawing buffer is `Rgba8Unorm`, not an sRGB framebuffer. The engine still asked for `ColorSpace::Linear`, so the rasterizer wrote linear values that the browser displayed as sRGB. The image went dark and local lights disappeared into the crush.

If the realized format is `Rgba8Unorm` or `Bgra8Unorm`, treat the surface as `ColorSpace::Srgb` so shaders encode.

### DIFF
--- a/blade-engine/src/lib.rs
+++ b/blade-engine/src/lib.rs
@@ -544,6 +544,22 @@ impl Engine {
         }
     }
 
+    fn color_space_for_surface(
+        requested: gpu::ColorSpace,
+        format: gpu::TextureFormat,
+    ) -> gpu::ColorSpace {
+        if format.is_srgb() {
+            requested
+        } else if matches!(
+            format,
+            gpu::TextureFormat::Rgba8Unorm | gpu::TextureFormat::Bgra8Unorm
+        ) {
+            gpu::ColorSpace::Srgb
+        } else {
+            requested
+        }
+    }
+
     fn hot_reload_if_needed(&mut self) {
         if !self.track_hot_reloads {
             return;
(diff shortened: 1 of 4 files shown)


blade-graphics/src/util.rs
```diff
@@ -97,6 +97,21 @@ impl super::TextureFormat {
         }
     }
 
+    /// True when storing this format converts linear values with the sRGB
+    /// transfer function. A non-sRGB 8-bit swapchain (WebGL's canvas) does
+    /// not, so the renderer has to encode.
+    pub const fn is_srgb(self) -> bool {
+        matches!(
```
kvark: no need to use matches! macro here

blade-engine/src/lib.rs
```diff
@@ -544,6 +544,22 @@ impl Engine {
         }
     }
 
+    fn color_space_for_surface(
```
kvark: This shouldn't be a responsibility of the engine. The contract is clear at the blade-graphics level: user specifies what color space their shaders are going to output, and the backend adjusts accordingly. User doesn't actually care if the low level surface is srgb or not.

At one of the higher levels (blade-render) this contract is specified even further: shaders output linear, period.

kvark: Agreed — that was the wrong layer. Engine always asks for `ColorSpace::Linear` (shaders output linear). WebGL present now encodes on the copy to the canvas, which is not an sRGB framebuffer.

blade-graphics/src/util.rs
kvark: Switched to a `match`.

kvark:
You're right — the canvas default FB is RGBA8, but present always blits from an offscreen we own. Native GLES already maps `ColorSpace::Linear` → `Rgba8UnormSrgb` so shaders write linear and the GPU encodes. Web now does the same (`cfe44bf`) instead of a present-time encode shader.

PR #380, opened 2026-08-30
Present WebGL surfaces with a texelFetch copy

kvark:
The offscreen is `SRGB8_ALPHA8` so shaders write linear and the GPU encodes. `blitFramebuffer` to the RGBA8 canvas decodes those bytes, and the compositor then treats the result as sRGB.

Replace the blit with a draw that `texelFetch`es the stored sRGB bytes (no filter, no gamma equation) onto the canvas. The canvas is always RGBA8, so this is the present path for every color space.

### DIFF
--- a/blade-graphics/src/gles/web.rs
+++ b/blade-graphics/src/gles/web.rs
@@ -1,10 +1,39 @@
 use glow::HasContext as _;
 use wasm_bindgen::JsCast;
 
+const PRESENT_VS: &str = r#"#version 300 es
+void main() {
+    vec2 pos[3] = vec2[3](vec2(-1.0, -1.0), vec2(3.0, -1.0), vec2(-1.0, 3.0));
+    gl_Position = vec4(pos[gl_VertexID], 0.0, 1.0);
+}
+"#;
+
+const PRESENT_FS: &str = r#"#version 300 es
+precision highp float;
+precision highp int;
+uniform sampler2D src;
+out vec4 frag;
+void main() {
+    // Same Y-flip as `present_blit`. texelFetch is a raw copy: no filter,
+    // no sRGB decode (that is part of texture filtering).
+    ivec2 size = textureSize(src, 0);
+    ivec2 p = ivec2(gl_FragCoord.xy);
+    p.y = size.y - 1 - p.y;
+    frag = texelFetch(src, p, 0);
+}
+"#;
+
+struct PresentCopy {
+    program: glow::Program,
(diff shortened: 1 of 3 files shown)


blade-graphics/src/gles/web.rs
```diff
@@ -1,20 +1,52 @@
...
     info: crate::SurfaceInfo,
     extent: crate::Extent,
+    color_space: crate::ColorSpace,
 }
 #[derive(Debug)]
 pub struct PlatformFrame {
     framebuf: glow::Framebuffer,
+    texture: glow::Texture,
     extent: crate::Extent,
+    color_space: crate::ColorSpace,
```
kvark: does it need to be a property of the frame? the context should already know what its swapchain color space is

blade-graphics/src/gles/web.rs
```diff
@@ -152,10 +221,72 @@ impl super::Context {
                 Some(surface.offscreen_texture),
                 0,
             );
+            gl.tex_parameter_i32(
```
kvark: should we loopify this? e.g. `for filter_target in [...] {...}`

blade-graphics/src/gles/web.rs
```diff
@@ -152,10 +221,72 @@ impl super::Context {
...
+                glow::TEXTURE_2D,
+                glow::TEXTURE_MIN_FILTER,
+                glow::NEAREST as i32,
+            );
+            gl.tex_parameter_i32(
+                glow::TEXTURE_2D,
+                glow::TEXTURE_MAG_FILTER,
+                glow::NEAREST as i32,
+            );
+            gl.tex_parameter_i32(
```
kvark: same here

kvark: Same for wrap: `TEXTURE_WRAP_S` / `TEXTURE_WRAP_T`.

blade-graphics/src/gles/web.rs
kvark: Looped `TEXTURE_MIN_FILTER` / `TEXTURE_MAG_FILTER`.

blade-graphics/src/gles/web.rs
kvark: Dropped it from the frame (and the surface). The WebGL canvas is always RGBA8, so present is always the `texelFetch` copy — no color-space branch.

kvark (commented):
unfortunate addition of 100 LOC but I don't see another way

PR #383, opened 2026-09-02
Transient buffer allocation

kvark:


blade-graphics/src/vulkan/descriptor.rs
```diff
@@ -61,6 +61,30 @@ impl DescriptorCounts {
     fn supports(self, required: Self) -> bool {
         self.max(required) == self
     }
+
+    fn max_sets_before_overflow(self) -> u32 {
```
kvark: I think we should just handle this transparently instead. When growing those, cap at `u32::MAX` each.

blade-graphics/src/lib.rs
```diff
@@ -357,6 +357,13 @@ pub struct BufferDesc<'a> {
     pub name: &'a str,
     pub size: u64,
     pub memory: Memory,
+    /// Prefer allocation optimized for short-lived or batch-lifetime buffers.
+    ///
+    /// This hint is independent of [`Memory`]: transient device, shared,
+    /// upload, and download buffers retain their normal visibility and access
+    /// properties. Backends without a distinct transient allocator may ignore
+    /// it.
+    pub transient: bool,
```
kvark: actually, I take it back. It should be in `Memory::DeviceTransient` instead. That would make it hurt the ergonomics much less, and I don't think we have a lot of valid cases where CPU-visible buffers need to be transient.

blade-graphics/src/vulkan/resource.rs
```diff
@@ -884,4 +897,19 @@ mod tests {
         );
         assert!(memory_allocate_flags(false).is_empty());
     }
+
+    #[test]
+    fn transient_is_independent_of_memory_location() {
```
kvark: let's remove this

PR #385, opened 2026-09-06
Prepare for blade-graphics-0.9 release

kvark:


blade-render/src/raster/mod.rs
```diff
@@ -1268,3 +1323,53 @@ fn mat4_transform(t: &gpu::Transform) -> glam::Mat4 {
     }
     .transpose()
 }
+
+#[cfg(test)]
+mod local_light_tests {
```
kvark: I don't think these tests are useful

PR #388, opened 2026-09-11
Preserve calibrated GPU timestamp ranges

kvark:


blade-graphics/src/lib.rs
```diff
@@ -1466,4 +1466,34 @@ pub struct Viewport {
...
+
+    /// Duration of each pass: this pass's start to the next start, or to
+    /// [`Self::done`] for the last pass.
+    pub fn pass_durations(&self) -> impl Iterator<Item = (&str, std::time::Duration)> + '_ {
+        self.passes
+            .iter()
+            .enumerate()
+            .filter_map(move |(i, (name, start))| {
+                let end = self.passes.get(i + 1).map(|(_, t)| *t).unwrap_or(self.done);
+                Some((name.as_str(), end.checked_duration_since(*start)?))
```
kvark: seems to me that we shouldn't expect `checked_duration_since` fail here, and should replace `filter_map` to `map`.

blade-graphics/src/vulkan/mod.rs
```diff
@@ -39,9 +39,100 @@ struct RayTracingDevice {
...
+fn calibrate_timestamps(
+    extension: CalibratedTimestampsExtension,
+    instance: &ash::Instance,
+    device: &ash::Device,
+) -> Option<TimestampCalibration> {
+    let info = [vk::CalibratedTimestampInfoKHR::default().time_domain(vk::TimeDomainKHR::DEVICE)];
+    let before = Instant::now();
+    let result = unsafe {
+        match extension {
+            CalibratedTimestampsExtension::Khr => {
```
kvark: Seems like this is just a promoted version of the EXT? in this case, we should just use EXT only for simplicity

blade-graphics/src/vulkan/mod.rs
```diff
@@ -677,6 +770,12 @@ impl crate::traits::CommandDevice for Context {
 
     fn submit(&self, encoder: &mut CommandEncoder) -> SyncPoint {
         let raw_cmd_buf = encoder.finish();
+        if let Some(ref timing) = encoder.device.timing {
+            let cmd_buf = encoder.buffers.first_mut().unwrap();
+            cmd_buf.timing_calibration = timing.calibrated_timestamps.and_then(|extension| {
```
kvark: do we know how much time it takes, roughly?

blade-graphics/src/vulkan/mod.rs
```diff
@@ -39,9 +39,100 @@ struct RayTracingDevice {
...
+            u64::MAX
+        } else {
+            (1u64 << bits) - 1
+        };
+        let raw_delta = timestamp.wrapping_sub(calibration.gpu_ticks) & mask;
+        let sign_bit = 1u64 << (bits - 1);
+        let delta_ticks = if raw_delta & sign_bit == 0 {
+            i128::from(raw_delta)
+        } else {
+            i128::from(raw_delta) - (1i128 << bits)
```
kvark: do we support both positive and negative difference here? why?
let's just assume the difference is positive and assert otherwise

blade-graphics/src/vulkan/command.rs
```diff
@@ -27,6 +27,54 @@ impl super::CrashHandler {
...
+            .unwrap();
+    }
+
+    let calibration =
+        cmd_buf
+            .timing_calibration
+            .take()
+            .unwrap_or_else(|| super::TimestampCalibration {
+                cpu: Instant::now(),
+                gpu_ticks: timestamps[n],
```
kvark: how is this going to work? if we didn't have a calibration, what's the plan?

blade-graphics/src/gles/command.rs
```diff
@@ -1,4 +1,16 @@
-use std::{str, time::Duration};
+use std::{
+    str,
+    time::{Duration, Instant},
+};
+
+fn map_gpu_ns(cal_cpu: Instant, cal_gpu_ns: u64, ts_ns: u64) -> Option<Instant> {
+    let delta = ts_ns as i128 - cal_gpu_ns as i128;
+    if delta >= 0 {
```
kvark: similar here, no need to make it overly general

blade-graphics/src/lib.rs
kvark: Done. `pass_durations()` now uses `map` and `duration_since`; a non-monotonic backend result is treated as an invariant violation instead of being silently dropped.

blade-graphics/src/vulkan/mod.rs
kvark: Done. Vulkan timing now requires and enables `VK_EXT_calibrated_timestamps`, and the device keeps only the EXT loader.

blade-graphics/src/vulkan/mod.rs
kvark: Measured on the RADV Strix system here, in a debug build with validation enabled: 572 per-submit calibration calls had a 24.2 µs median, 26.8 µs mean, 38.7 µs p95, and a 21.3–75.3 µs range. The call is now made under the queue lock immediately before submission.

blade-graphics/src/vulkan/mod.rs
kvark: Done: conversion is forward-only (while still handling `timestampValidBits` wrap) and asserts on a negative interval.

That assertion also explained the observed 11.85 s gap: it is a GFX11.5 amdgpu bug, not a real negative interval or a Blade scaling error. On this RADV STRIX1 machine, the two raw values differ by about 1.183 billion ticks; at the reported 10.019 ns period that is 11.8515 s, and the offset is stable across runs. AMD independently reported the same Vulkan CTS failure on Strix/Krackan: calibrated timestamps read `GOLDEN_TSC_COUNT`, while `vkCmdWriteTimestamp` uses `CP_MES_MTIME`; the clocks have the same frequency but different epochs. Their proposed kernel fix switches GFX11.5+ calibration to `CP_MES_MTIME`: https://www.spinics.net/lists/amd-gfx/msg141414.html and https://www.spinics.net/lists/amd-gfx/msg141433.html. Vulkan requires the device-domain value to be comparable with `vkCmdWriteTimestamp`: https://docs.vulkan.org/refpages/latest/refpages/source/VkTimeDomainKHR.html

blade-graphics/src/vulkan/command.rs
kvark: Removed the fallback. Vulkan timing is now disabled unless `VK_EXT_calibrated_timestamps` is advertised, and every timed submission must carry a real calibration; resolution simply unwraps that invariant.

blade-graphics/src/gles/command.rs
kvark: Done. GLES conversion is now forward-only and asserts if a query predates its calibration. I also bracketed the `GL_TIMESTAMP` read and use the CPU midpoint as its anchor.

kvark: do we have to support negative here?

blade-graphics/src/metal/command.rs
```diff
@@ -4,7 +4,22 @@ use objc2_metal::{
...
+    marker::PhantomData,
+    mem,
+    ptr::NonNull,
+    slice,
+    time::{Duration, Instant},
+};
+
+fn map_gpu_ns(cal_cpu: Instant, cal_gpu_ns: u64, ts_ns: u64) -> Option<Instant> {
+    let delta = ts_ns as i128 - cal_gpu_ns as i128;
+    if delta >= 0 {
```
kvark: and here?

blade-graphics/src/vulkan/mod.rs
```diff
@@ -39,9 +39,100 @@ struct RayTracingDevice {
...
+
+#[derive(Clone)]
 struct TimingDevice {
     period: f32,
+    valid_bits: u32,
+    calibrated_timestamps: Option<CalibratedTimestampsExtension>,
+}
+
+impl TimingDevice {
+    fn map_timestamp(&self, calibration: TimestampCalibration, timestamp: u64) -> Option<Instant> {
```
kvark: I don't understand this optionality. Seems like we are overly defensive?
we should strictly find out if timestamps work either at init() time or at the start of submit. Any processing of the timestamps should just assume they are correct

blade-graphics/src/vulkan/mod.rs
kvark: I still see this code supporting negatives? For this bug on our machine, can't we detect it earlier?

blade-graphics/src/vulkan/mod.rs
kvark: I still see `CalibratedTimestampsExtension::Khr` variant, why do we need it?

blade-graphics/src/lib.rs
```diff
@@ -1466,4 +1466,52 @@ pub struct Viewport {
...
```
kvark: it doesn't look like your changes got applied at all

blade-graphics/src/traits.rs
```diff
@@ -66,7 +66,15 @@ pub trait CommandEncoder {
     fn start(&mut self);
     fn init_texture(&mut self, texture: Self::Texture);
     fn present(&mut self, frame: Self::Frame);
+    /// Resolve timestamp queries for all command buffers in this encoder.
+    ///
+    /// The caller must first wait for its latest submission to complete.
+    fn resolve_timings(&mut self);
```
kvark: why is this a separate method from `timings()` and `timing_spans`? we should have one method that resolves and gives us the results

blade-graphics/src/vulkan/mod.rs
kvark: Removed in `69cddf6`. The enum and KHR path are gone; Vulkan timing now requires/enables `VK_EXT_calibrated_timestamps` and stores only the EXT loader.

blade-graphics/src/vulkan/mod.rs
kvark: Addressed in `69cddf6`. Mapping now accepts only a short forward modular delta and asserts otherwise. More importantly, device initialization now submits one timestamp query bracketed by two calibrated samples; the broken clock relationship on this machine is detected there and timing is disabled before encoders are created.

blade-graphics/src/metal/command.rs
kvark: No. Removed in `69cddf6`: Metal now brackets the queries with two clock samples, asserts both forward bounds, and interpolates within that interval.

blade-graphics/src/lib.rs
kvark: You were right about the visible branch state: the earlier replies preceded the actual changes, and the intervening force-push left the old implementation in the PR. `69cddf6` is now on the branch; the old `Timings`/`pass_durations` code and duplicated timing accessors are removed.

blade-graphics/src/traits.rs
kvark: Collapsed in `69cddf6`. The trait now has one method, `timings(&mut self) -> &[GpuTimingSpan]`; it resolves available Vulkan queries and returns the results. `resolve_timings` and `timing_spans` are gone.

blade-graphics/src/gles/command.rs
kvark: No. Removed in `69cddf6`: GLES now asserts that the query is at or after calibration and subtracts directly.

blade-graphics/src/vulkan/mod.rs
kvark: Simplified in `69cddf6`. Timestamp mapping no longer returns `Option`; initialization establishes the clock invariant with a real submitted query, and normal resolution directly maps timestamps under that invariant.

blade-graphics/src/traits.rs
```diff
@@ -66,7 +66,9 @@ pub trait CommandEncoder {
     fn start(&mut self);
     fn init_texture(&mut self, texture: Self::Texture);
     fn present(&mut self, frame: Self::Frame);
-    fn timings(&self) -> &super::Timings;
+    /// Resolve available timestamp queries and return their pass ranges on the
+    /// process monotonic clock.
+    fn timings(&mut self) -> &[super::GpuTimingSpan];
```
kvark: if it's "&mut self" then the method should be called with an action, e.g. "get_timings()"

blade-graphics/src/lib.rs
```diff
@@ -1466,4 +1466,16 @@ pub struct Viewport {
     pub depth: std::ops::Range<f32>,
 }
 
-pub type Timings = Vec<(String, std::time::Duration)>;
+/// A timestamped GPU pass mapped onto the process monotonic clock.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct GpuTimingSpan {
+    pub name: String,
+    pub start: std::time::Instant,
```
kvark: why did you fall back to the spans? do we expect passes to be executed out of order?
we insert pretty much full barriers between passes anyway, so no need to assume they will overlap
besides, when registering the perf scopes the tools don't always appreciate overlaps anyway. It's more natural to look at it as a sequence strictly.

blade-graphics/src/vulkan/mod.rs
```diff
@@ -39,9 +43,158 @@ struct RayTracingDevice {
...
+        );
+        TimestampCalibration {
+            cpu: before + call_duration.div_f64(2.0),
+            gpu_ticks: timestamps[0],
+            period: self.period,
+            valid_bits: self.valid_bits,
+        }
+    }
+
+    fn validate(&self, device: &ash::Device, queue: vk::Queue, queue_family_index: u32) -> bool {
```
kvark: do we really need that? a lot of code here. Blade is trying to be very minimal

blade-graphics/src/lib.rs
kvark: Agreed. `5df4778` removes `GpuTimingSpan` and restores one strictly ordered `Timings` sequence: pass start times plus the final completion time. `pass_durations()` ends each pass at the next pass start (and the last at `done`), so consumers don't produce overlapping scopes.

blade-graphics/src/vulkan/mod.rs
kvark: No, it was too much machinery for Blade. Removed in `5df4778`—including the throwaway command pool, command buffer, query pool, and blocking initialization submission. The normal forward timestamp assertions remain, while the known GFX11.5 clock bug stays a driver concern.

blade-graphics/src/traits.rs
kvark: Renamed in `5df4778`: the trait, all backends, and callers now use `get_timings()`.

blade-graphics/src/metal/mod.rs
```diff
@@ -716,19 +743,24 @@ impl crate::traits::CommandDevice for Context {
...
         }
     }
 
     fn destroy_command_encoder(&self, _command_encoder: &mut CommandEncoder) {}
 
     fn submit(&self, encoder: &mut CommandEncoder) -> SyncPoint {
         use metal::MTLCommandBuffer as _;
+        if let Some(ref mut td_array) = encoder.timing_datas {
+            let td = td_array.first_mut().unwrap();
+            td.calibration = Some(sample_timestamps(&encoder.device));
```
kvark: why do we need "encoder.device" at all? We have "&self" as the context, and the context owns the device already.

blade-graphics/src/vulkan/command.rs
```diff
@@ -800,7 +809,20 @@ impl crate::traits::CommandEncoder for super::CommandEncoder {
...
+        self.timings.passes.clear();
+        let mut done = None;
+        for cmd_buf in self.buffers.iter_mut() {
+            if let Some(candidate) =
+                resolve_timing_buffer(&self.device.core, cmd_buf, &mut self.timings)
+            {
+                done = Some(done.map_or(candidate, |previous: Instant| previous.max(candidate)));
+            }
+        }
+        self.timings.passes.sort_unstable_by_key(|entry| entry.1);
```
kvark: why are we sorting the passes at all?

blade-graphics/src/vulkan/command.rs
```diff
@@ -800,7 +809,20 @@ impl crate::traits::CommandEncoder for super::CommandEncoder {
         });
     }
 
-    fn timings(&self) -> &crate::Timings {
+    fn get_timings(&mut self) -> &crate::Timings {
+        self.timings.passes.clear();
+        let mut done = None;
+        for cmd_buf in self.buffers.iter_mut() {
+            if let Some(candidate) =
+                resolve_timing_buffer(&self.device.core, cmd_buf, &mut self.timings)
```
kvark: could make it a non-public method on `CommandBuffer`

blade-graphics/src/vulkan/command.rs
```diff
@@ -1370,8 +1392,34 @@ impl crate::traits::RenderPipelineEncoder for super::PipelineEncoder<'_, '_> {
...
+            valid_bits: 8,
+        };
+
+        assert_eq!(calibration.map(5), cpu + Duration::from_nanos(22));
+    }
 
-    use super::super::{CommandEncoder, PassKind, PassKinds};
+    #[test]
+    #[should_panic(expected = "predates calibration")]
+    fn calibrated_timestamp_mapping_rejects_backward_time() {
```
kvark: not very useful, remove

kvark: Removed in `59306b3`.

blade-graphics/src/vulkan/command.rs
kvark: Done in `59306b3`: the resolver is now the private `CommandBuffer::resolve_timings()` method.

blade-graphics/src/vulkan/command.rs
kvark: Removed in `59306b3`. The rotated command-buffer ring is now traversed directly in submission order (older buffers, then the current buffer), so passes are appended in order and no timestamp sort is needed.

blade-graphics/src/metal/mod.rs
kvark: Removed in `59306b3`. `CommandEncoder` no longer retains a device: `submit()` samples through the context's device, and timing resolution gets the device from the encoder's existing command queue.

PR #390, opened 2026-09-12
Harden directional shadows for skinned receivers

kvark:
## Summary
- Skinned receivers (Quaternius mechs in geofront) went pure-black under lavapipe when the directional shadow map was on: shadow-map texels (~0.27 m at the old 512/70 defaults) dwarfed `normal_bias` 0.06, `Depth32Float` pipeline depth bias is nearly a no-op on software rasterizers, and strength 0.88 left no dusk ambient fill.
- Offset receivers along both the shading normal (slope-scaled by N·L) and the light direction; raise shadow-pass slope bias; safer defaults (1024², strength 0.72, dual world-space biases). PCF texel size comes from `textureDimensions` so `shadow_params.w` carries `depth_bias`.

## Test plan
- [ ] `cargo check -p blade-render`
- [ ] geofront lavapipe/Xvfb with shadows re-enabled: mechs stay colored, building/contact shadows visible, AUTOPLAY VICTORY
- [ ] Optional: ignored `snapshot_animated_skin` GPU test still links

### DIFF
--- a/blade-render/code/raster.wgsl
+++ b/blade-render/code/raster.wgsl
@@ -30,7 +30,7 @@ struct RasterFrameParams {
     ambient_color: vec4<f32>,
     // x: environment map enabled, y: the surface needs sRGB encoding
     settings: vec4<f32>,
-    // x: enabled, y: strength, z: receiver normal bias, w: texel size
+    // x: enabled, y: strength, z: receiver normal bias, w: light-dir bias
     shadow_params: vec4<f32>,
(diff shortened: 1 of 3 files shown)


kvark:
Follow-up commit `221e982` centers the ortho on the look-at and eases contact bias. Geofront re-enable: https://github.com/kvark/geofront/pull/11 (AUTOPLAY VICTORY under lavapipe).

PR #394, opened 2026-09-16
Resolve GPU timings for the submission just waited on

kvark:
Follow-up to the `[P1, still open]` review note on kvark/meganeura#187 (`discussion_r4021541144`), which pointed at Blade rather than Meganeura: Metal's timing contract does not match an immediate read after `wait`.

Rewrites the timings code in all backends. Forces the user to wait for GPU themselves, and only obtain the timings for the last submitted command buffer.


blade-graphics/src/metal/command.rs
```diff
@@ -487,6 +528,23 @@ impl crate::traits::CommandEncoder for super::CommandEncoder {
...
+        if let Some(ref mut td_array) = self.timing_datas {
+            let mut done = None;
+            // `start` rotates the slot being recorded to index zero, so the
+            // submitted ones run oldest-first from index one and wrap back
+            // around to index zero, which holds the latest submission until
+            // the next `start` takes it over.
+            let (current, older) = td_array.split_at_mut(1);
+            for td in older.iter_mut().chain(current.iter_mut()) {
+                if let Some(candidate) = td.resolve_timings(&self.queue, &mut self.timings) {
+                    done = Some(candidate);
```
kvark: This is a bit dangerous. We are assigning the done flag based on the gpu completing some of the command buffers? We may end up returning the done time to the user even though some commands are still in flight

blade-graphics/src/metal/command.rs
```diff
@@ -487,6 +528,23 @@ impl crate::traits::CommandEncoder for super::CommandEncoder {
...
 
     fn get_timings(&mut self) -> &crate::Timings {
+        self.timings.passes.clear();
+        if let Some(ref mut td_array) = self.timing_datas {
+            let mut done = None;
+            // `start` rotates the slot being recorded to index zero, so the
+            // submitted ones run oldest-first from index one and wrap back
+            // around to index zero, which holds the latest submission until
+            // the next `start` takes it over.
+            let (current, older) = td_array.split_at_mut(1);
```
kvark: Actually why are we reporting timings from all of the past work? Shouldn't we take the last submit only? Wouldn't that be matching Vulkan?

blade-graphics/src/metal/command.rs
kvark: Fixed in f478feb.

You're right that this is wrong, though the mechanism is slightly different from what it looks like. `done` only ever came from a slot that *did* resolve, so it was never a completion time for in-flight work directly. The real damage was next to it: because each resolve **appended** to `timings.passes`, two finished submissions produced one flat pass list that nobody executed, and `pass_durations` chains `passes[i] -> passes[i+1]`, so the idle gap between the two submissions got charged to whichever pass sat on the boundary. `done` was then the newest one's, describing only the tail of that merged list.

Each resolve now replaces rather than appends, so `passes` and `done` always describe the same single submission. In-flight slots still resolve to nothing and cannot contribute a `done`.

---
_Generated by [Claude Code](https://claude.ai/code)_

blade-graphics/src/metal/command.rs
kvark: Agreed, and done in f478feb - but one correction on the premise: **Vulkan was not doing that either.** It accumulates exactly the same way today on `main`:

```rust
let (current, older) = self.buffers.split_at_mut(1);
for cmd_buf in older.iter_mut().chain(current.iter_mut()) {
    if let Some(candidate) = cmd_buf.resolve_timings(&self.device.core, &mut self.timings) {
        done = Some(candidate);   // passes are pushed, not replaced
    }
}
```

I mirrored that loop into Metal deliberately, to make the two backends agree. So "take the last submit only" is a change to both, and I've applied it to both - fixing only Metal would have reintroduced the divergence this PR exists to remove.

On "the last submit" specifically: I read that as *the newest submission that has completed*, not literally the most recent `submit`. Those differ in the `FramePacer` flow, which submits frame N and then waits on frame N-1, so at `get_timings` time the newest submission is usually still in flight. Reporting literally-the-last would hand the egui overlay an empty list every frame. Newest-completed keeps that working and still gives a profiler the submission it just waited on.

`timings_report_only_the_newest_completed_submission` covers it: two submissions finish between reads, only the second is reported. I confirmed it fails against the old accumulate behaviour (`["stale-pass", "fresh-pass"]`) before fixing it.

---
_Generated by [Claude Code](https://claude.ai/code)_

blade-graphics/src/metal/mod.rs
kvark: should this use `RawCommandBuffer`?

blade-graphics/src/metal/mod.rs
```diff
@@ -219,6 +219,10 @@ struct TimingData {
     pass_names: Vec<String>,
     sample_buffer: Retained<ProtocolObject<dyn metal::MTLCounterSampleBuffer>>,
     calibration: Option<TimestampSample>,
+    /// The submission that wrote `sample_buffer`, held until its counters are
+    /// harvested. Metal leaves the sample buffer undefined until the command
+    /// buffer completes, so this is what tells us a read is safe.
+    cmd_buf: Option<RawCommandBuffer>,
```
kvark: I don't think we need it here. Would be simpler to just store the last submitted one on `CommandEncoder` itself

blade-graphics/src/vulkan/command.rs
```diff
@@ -815,6 +824,11 @@ impl crate::traits::CommandEncoder for super::CommandEncoder {
         self.timings.passes.clear();
         let mut done = None;
         // `start` rotates the current buffer to index zero, leaving older submissions in order.
+        //
+        // Walking them oldest-first leaves the newest completed one reported:
```
kvark: why are we walking at all? why not just looking at the last submitted one?

blade-graphics/src/metal/mod.rs
kvark: Yes - done in c0f02e2.

---
_Generated by [Claude Code](https://claude.ai/code)_

blade-graphics/src/metal/mod.rs
kvark: Done in c0f02e2 - `TimingData::cmd_buf` is gone, the encoder holds `last_submission` instead, and `start` drops it (rotating the slots is exactly where those counters stop being the ones at index zero).

---
_Generated by [Claude Code](https://claude.ai/code)_

blade-graphics/src/vulkan/command.rs
kvark: Switched to the last submitted one only, in c0f02e2 - both backends.

**First, a correction to what I told you earlier in this PR.** I claimed reporting only the last submission would "hand the egui overlay an empty list every frame", because `FramePacer` submits frame N and then waits on frame N-1, leaving N in flight. I went and measured it instead of continuing to assert it, and I was wrong. Reproducing the pacer ordering on lavapipe with a deliberately heavy dispatch:

```
frame 0: newest_complete=false  reported=["frame-0"]  done_age=0.2ms
frame 1: newest_complete=false  reported=["frame-1"]  done_age=0.1ms
...
```

The timeline semaphore for frame N has genuinely not signalled, but the query pool results are already in, and `done_age` (how far `done` sits behind `Instant::now()`) is a fraction of a millisecond - so it is that frame's own data, not a stale read. I checked that specifically, because `cmd_reset_query_pool` is recorded *inside* the command buffer, so a buffer that has not started executing still holds available results from its previous use. Not what is happening here.

So the objection I raised does not hold, and `buffers[0]` is enough. The walk is gone.

The one thing I could not measure from here is a pipelined discrete GPU, where frame N may still be executing when the overlay reads. There the result is empty for that frame rather than the previous frame's numbers. That is a visible difference in the timing panel, not a correctness problem, and it is now what the trait documents - flagging it in case you would rather the overlay fall back.

---
_Generated by [Claude Code](https://claude.ai/code)_

blade-graphics/src/metal/mod.rs
```diff
@@ -748,19 +760,27 @@ impl crate::traits::CommandDevice for Context {
...
     fn submit(&self, encoder: &mut CommandEncoder) -> SyncPoint {
         use metal::MTLCommandBuffer as _;
-        if let Some(ref mut td_array) = encoder.timing_datas {
-            let td = td_array.first_mut().unwrap();
-            td.calibration = Some(sample_timestamps(&self.device.lock().unwrap()));
-        }
+        let timed = match encoder.timing_datas {
+            Some(ref mut td_array) => {
+                let td = td_array.first_mut().unwrap();
+                td.calibration = Some(sample_timestamps(&self.device.lock().unwrap()));
```
kvark: should we do the same thing - just store it on the `CommandEncoder` itself?

blade-graphics/src/metal/mod.rs
```diff
@@ -748,19 +760,27 @@ impl crate::traits::CommandDevice for Context {
...
+                !td.pass_names.is_empty()
```
kvark: similarly, this

tests/gpu_examples.rs
```diff
@@ -308,6 +308,157 @@ fn manual_barrier_gpu_test() {
...
+///
+/// `Timings` describes one submission: `passes` is a single run in execution
+/// order and `done` is where that run ended. Concatenating two of them would
+/// invent a pass list nobody executed, and `pass_durations` would report the
+/// idle gap between submissions as time spent inside a pass. Only the last
+/// submission is reported, so the earlier one is dropped rather than appended.
+#[cfg(not(gles))]
+#[test]
+#[ignore = "requires a working GPU context"]
+fn timings_report_only_the_last_submission() {
```
kvark: can we merge into a single small test? Also, why excluding gles? may just bail out if the device doesn't support timing

tests/gpu_examples.rs
```diff
@@ -308,6 +308,157 @@ fn manual_barrier_gpu_test() {
...
+    context.sync_buffer(input.into(), input.size(), gpu::BufferTarget::Data);
+
+    let shader = context.create_shader(gpu::ShaderDesc {
+        source: include_str!("shaders/dispatch.wgsl"),
+        naga_module: None,
+    });
+    let global_layout = DispatchGlobals::layout();
+    let mut pipeline = context.create_compute_pipeline(gpu::ComputePipelineDesc {
+        name: "timing-test",
+        data_layouts: &[&global_layout],
```
kvark: curious why we couldn't just have empty data layout slice

blade-graphics/src/metal/mod.rs
kvark: Done in 887be51 - and it went further than just the calibration, since your next comment applies to `pass_names` too.

With only the last submission ever reported there is one set of pass names and one calibration in play, so neither needs to exist per sample buffer. `TimingData` is gone; the encoder holds a `TimingState` with the names, the calibration, the submission waiting to be harvested, and the sample buffers.

The buffers stay a ring - recording the next submission must not overwrite counters an in-flight one has yet to have read out of it - but rotating them is precisely what gives up the previous submission's counters, so `begin_recording` now does both in one place instead of leaving those two facts apart.

---
_Generated by [Claude Code](https://claude.ai/code)_

blade-graphics/src/metal/mod.rs
kvark: Yes - `pass_names` moved too, in 887be51. Same reasoning as the calibration; see the reply above.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/gpu_examples.rs
kvark: Merged into one test in 887be51: `get_timings_reports_the_last_submission`. It runs four submissions labelled differently and reads after three of them, so one pass covers both halves of the contract - each read sees the submission it waited on, and the unread one is dropped rather than merged into the next. Confirmed it fails both ways it should:

- `get_timings` reverted to a plain getter -> `[[], [], []]`
- the walk reintroduced -> `[["first"], ["second"], ["last", "unread"]]`

**On GLES: a capability bail does not cover it.** I checked rather than assuming - a GLES context here reports `capabilities().timing == true`, because the device does have `GL_EXT_disjoint_timer_query`. The lag is in the backend, not the device: GLES reads a submission's queries while submitting the *next* one, so it is a frame behind by construction, and the test would run and fail rather than bail.

It is that way because `get_timings` has no GL context to read with - only `submit` does, and the queries for the submission being made have not been issued yet when `finish` runs. Resolving them after the command loop instead would work, but `glGetQueryObjectui64v(GL_QUERY_RESULT)` blocks, so it would turn every submit into a `glFinish`. That seemed like a separate call to make, so I left the behaviour alone, documented it on the trait, and said in the test why the exclusion is about the backend rather than the device. Happy to take the stall if you would rather the backends agree.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/gpu_examples.rs
kvark: No reason - just inherited from `run_dispatch_gpu_test`, which I copied the setup from. Fixed in 887be51: the test now uses an inline `@compute @workgroup_size(1) fn main() {}` with `data_layouts: &[]`, which drops both buffers, the `sync_buffer`, and the `DispatchGlobals` plumbing. Nothing needs to be computed here, only timed. That is most of why the test file shrinks by ~120 lines.

---
_Generated by [Claude Code](https://claude.ai/code)_

blade-graphics/src/metal/command.rs
```diff
@@ -487,6 +527,24 @@ impl crate::traits::CommandEncoder for super::CommandEncoder {
...
+            match timing.last_submission_status() {
+                Some(metal::MTLCommandBufferStatus::Completed) => {
+                    timing.last_submission = None;
+                    if let Some(done) = timing.resolve_timings(&self.queue, &mut self.timings) {
+                        self.timings.done = done;
+                    }
+                }
+                // A failed submission never wrote its counters, so there is
+                // nothing to come back for.
+                Some(metal::MTLCommandBufferStatus::Error) => timing.forget(),
```
kvark: why do we need to do it here? let's try to make `get_timings` idempotent

blade-graphics/src/metal/command.rs
```diff
@@ -487,6 +527,24 @@ impl crate::traits::CommandEncoder for super::CommandEncoder {
...
+        self.timings.passes.clear();
+        // Only the last submission is reported, and only once it has finished.
+        // `start` has not run since it was submitted - that gives the counters
+        // up - so they are still the sample buffer at index zero.
+        if let Some(ref mut timing) = self.timing {
+            match timing.last_submission_status() {
+                Some(metal::MTLCommandBufferStatus::Completed) => {
+                    timing.last_submission = None;
+                    if let Some(done) = timing.resolve_timings(&self.queue, &mut self.timings) {
+                        self.timings.done = done;
```
kvark: this looks weird, couldn't `resolve_timings` assign this?

blade-graphics/src/metal/command.rs
```diff
@@ -487,6 +527,24 @@ impl crate::traits::CommandEncoder for super::CommandEncoder {
     }
 
     fn get_timings(&mut self) -> &crate::Timings {
+        self.timings.passes.clear();
+        // Only the last submission is reported, and only once it has finished.
+        // `start` has not run since it was submitted - that gives the counters
+        // up - so they are still the sample buffer at index zero.
+        if let Some(ref mut timing) = self.timing {
```
kvark: a bit of confusion between `self.timing` and `self.timings`

blade-graphics/src/metal/command.rs
```diff
@@ -487,6 +527,24 @@ impl crate::traits::CommandEncoder for super::CommandEncoder {
...
+            match timing.last_submission_status() {
+                Some(metal::MTLCommandBufferStatus::Completed) => {
+                    timing.last_submission = None;
```
kvark: why do we need to do it? again, let's try to be idempotent

blade-graphics/src/metal/mod.rs
```diff
@@ -755,12 +775,20 @@ impl crate::traits::CommandDevice for Context {
...
+                !timing.pass_names.is_empty()
+            }
+            None => false,
+        };
         let cmd_buf = encoder.finish();
         cmd_buf.commit();
+        if let Some(ref mut timing) = encoder.timing {
+            // Remember the submission so `get_timings` knows when the counters
+            // it wrote become readable. One that timed nothing has none to give.
+            timing.last_submission = timed.then(|| cmd_buf.clone());
```
kvark: can we avoid fancy constructs? just writing `if timed { Some(cmd_buf.clone()) } else { None }` would be not much longer but using the basic primitives

blade-graphics/src/metal/command.rs
```diff
@@ -197,12 +199,82 @@ impl<'a, const N: crate::ResourceIndex> crate::ShaderBindable
...
+    /// Whatever the previous one left behind goes with it - it is only still
+    /// here if the caller never asked for its timings - along with the claim
+    /// that sample buffer zero has something to harvest.
+    fn begin_recording(&mut self) {
+        self.sample_buffers.rotate_left(1);
+        self.forget();
+    }
+
+    /// Give up on the last submission's counters without reading them.
+    fn forget(&mut self) {
```
kvark: we shouldn't need this on a separate method. We should just do it inside `begin_recording` only

blade-graphics/src/metal/command.rs
kvark: Gone in 783909f, along with the whole `match`. Since a failed submission never leaves `Completed`, `resolve_timings` already declines to touch it, and `begin_recording` clears it at the next `start`. There was nothing for `get_timings` to do about a failure, so it no longer tries.

---
_Generated by [Claude Code](https://claude.ai/code)_

blade-graphics/src/metal/command.rs
kvark: Yes - it does now, in 783909f. `resolve_timings` writes `passes` and `done` together and returns `()`, instead of filling one and handing the other back for the caller to assign. Same on the Vulkan side.

---
_Generated by [Claude Code](https://claude.ai/code)_

blade-graphics/src/metal/command.rs
kvark: Fair - one letter apart and meaning different things. Renamed the state to `timing_state` in 783909f; `timings` stays as the resolved result, which is what the other backends call it.

---
_Generated by [Claude Code](https://claude.ai/code)_

blade-graphics/src/metal/command.rs
kvark: Done in 783909f - `get_timings` is idempotent now, and this line was most of why it wasn't.

Resolving consumed the very state that said there was something to resolve, so a second read fell through to an empty set. The fix was to stop clearing on the way out and let `timings` hold the last resolve: draining `pass_names` is already enough to keep a second call from redoing the work, and it leaves the answer standing. `last_submission` can just stay set.

That collapsed the rest of the review round into it - no status `match`, no `forget`, no `Option<Instant>` threaded back to the caller, and `submit` no longer has to decide whether a submission is worth remembering, because `resolve_timings` filters the empty ones anyway.

The test reads twice at the end and asserts the second read matches the first. Confirmed it fails without the fix: `repeated: []` against `["last"]`.

---
_Generated by [Claude Code](https://claude.ai/code)_

blade-graphics/src/metal/mod.rs
kvark: Done in 783909f, and the construct turned out not to be needed at all rather than needing rewriting.

It existed to avoid remembering a submission that timed nothing. But `resolve_timings` returns early on empty `pass_names` regardless, so that case was already handled one level down - `submit` now just stores `Some(cmd_buf.clone())` and the `timed` flag is gone with it.

---
_Generated by [Claude Code](https://claude.ai/code)_

blade-graphics/src/metal/command.rs
kvark: Gone in 783909f - inlined into `begin_recording`, which is now the only place that gives those counters up.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #397, opened 2026-09-19
Allow opt-in serialized Vulkan command replay

kvark:
Add `CommandEncoder::start_reusable` and `Context::try_replay` for workloads whose command sequence and resource bindings stay fixed while buffer contents change.

- Vulkan records without `ONE_TIME_SUBMIT` only when explicitly requested and can resubmit after the caller waits for completion.
- Ordinary recording keeps its existing usage flags. Starting another recording invalidates replay.
- Timed encoders, presentation, Metal, and GLES return the ordinary-recording fallback; no hidden re-recording or timing substitution.
- Extend the existing dispatch/manual-barrier/timestamp tests rather than adding another GPU test binary.

Motivation: a matched SmolLM2 Vulkan diagnostic spends about 1.3–1.6 ms per decode step recording commands on the host. A serialized replay prototype reduced RTX 5070 decode from 3.21 to 1.81 ms and Arc B570 from 5.64 to 4.38 ms without changing kernels or logits validation. Meganeura integration explicitly invalidates on plan/binding/profiling changes; training remains unchanged.

Validation: `cargo check -p blade-graphics`, Clippy with warnings denied, formatting. The three affected GPU tests pass on NVIDIA, Intel and lavapipe, also with Khronos validation enabled and no validation messages. Other backends still need CI.

No raw data or binary artifacts. Prototype source is preserved separately at `experiment/llama-replay-2026-09-19`.

[Meganeura integration and matched results](https://github.com/kvark/meganeura/pull/192). All Blade CI jobs passed in [run 1111](https://github.com/kvark/blade/actions/runs/35462108460).


kvark:
I don't think we want this in Blade at this point. Saving 1ms on command encoding (while using lots of passes and dispatches) is only a win on some very marginal use cases.

kvark:
Agreed. I am withdrawing this API and removing the Meganeura integration from #192. The source remains on `perf/compute-command-replay` at `4befad4f5fbd427c1aca4b1b001fb8b7acd8e109` for the attribution mini-study; no binary artifacts are needed. The observed gain is CPU command recording on a small batch-one decode workload, not a GPU-kernel optimization.

PR #398, opened 2026-09-19
Sample GPU timestamps at the pass barrier's source stage

kvark:
A `TOP_OF_PIPE` query is not ordered by a compute-to-compute barrier, so the previous pass's work can be charged to the next label. On B570 that mis-charge was the whole heavy pass: 0.104 µs on its own label and about 645 µs on the following light label.

The timestamp at a pass boundary now uses the source stage of the barrier recorded there. Compute followed by compute is sampled at `COMPUTE_SHADER`, which is the completion that releases the next pass. `vkCmdWriteTimestamp` accepts one stage flag, so a source that names several stages is sampled at `ALL_COMMANDS`. A pass opened with no preceding barrier stays at `TOP_OF_PIPE`.

Rebased onto current `main` and squashed to one commit. The earlier full-pipe `BOTTOM_OF_PIPE` experiment is not what this commit does.


blade-graphics/src/vulkan/command.rs
```diff
@@ -344,9 +344,11 @@ impl super::CommandEncoder {
             }
             let query_pool = timing.query_pool;
             unsafe {
+                // Wait for preceding work before starting this pass's interval.
+                // TOP_OF_PIPE can charge that work to the following pass.
```
kvark: this doesn't look right. This call is going to be used for both the start of the scopes and the finish.
TOP_OF_PIPE works for the starts, and it will work well for the first scope. Nothing bad with just keeping it here. We aren't going to get the exact timing for each pass, perhaps, but is it a problem? I'd imagine it's only a problem on some very small passes. If it's really something that we have to solve, we should be issuing both begin and end timestamps for each pass, but that would increase the overhead.

blade-graphics/src/vulkan/command.rs
```diff
@@ -1443,6 +1461,28 @@ mod tests {
         );
     }
 
+    #[test]
+    fn timestamp_stage_matches_the_barrier_source() {
```
kvark: not useful, let's remove

blade-graphics/src/vulkan/command.rs
```diff
@@ -517,18 +517,34 @@ impl super::CommandEncoder {
...
+    }
+
     /// Insert a global memory barrier at the current pass boundary.
     ///
     /// The source covers all passes recorded since the previous barrier. A
     /// caller-placed barrier remains immediate and uses a conservative
     /// destination because its consumer has not been declared yet.
     pub fn barrier(&mut self) {
-        self.barrier_before(super::PassKind::Unknown);
+        if let Some(source) = self.barrier_before(super::PassKind::Unknown) {
```
kvark: should this be just `self.boundary_source = self.barrier_before(...);`?

PR #399, opened 2026-09-21
Expose cooperative matrix shapes and compute shared-memory capacity

kvark:
Report the actual `[M, N, K]` cooperative shapes instead of collapsing support to a square tile size. B570 exposes `[8, 16, 16]`; the old query hid it entirely. Also expose the subgroup width and maximum compute workgroup shared memory so callers can reject illegal kernel layouts before compilation.

- Keep only subgroup-scoped, non-saturating f32/f32 and f16/f32 configurations representable by Naga's current 8/16-dimensional IR.
- Preserve the Metal 8x8x8 f32 capability; report Metal's threadgroup-memory limit. GLES reports an unknown limit as zero.
- Update the existing square-matmul example to choose only a shape its generator supports. Capability discovery does not imply automatic rectangular codegen.

This changes the `CooperativeMatrix` API (`f32_tile`/`f16_tile` become shape vectors), without adding dependencies, a kernel policy, test executables, or benchmark artifacts. BF16 and NV cooperative_matrix2 are not represented by this change.

Checked with `cargo check -p blade-graphics --lib --examples --offline`. Direct Vulkan capability queries agree with the exposed lists on RTX 5070 and Arc B570; independent rectangular-matmul diagnostics validate full outputs on both. Metal/GLES are updated but not executed locally. `cargo clippy -p blade-graphics --lib --offline -- -D warnings`, all nine library tests, and `cargo check -p blade --example matmul --offline` also pass.

The timestamp boundary fix remains separate in #398.


blade-graphics/src/lib.rs
```diff
@@ -236,27 +236,32 @@ pub struct MemoryStats {
...
+/// An empty list means the corresponding operand type is unsupported.
+/// Only dimensions representable by Naga's cooperative types are included.
+#[derive(Clone, Debug, Default, PartialEq)]
 pub struct CooperativeMatrix {
-    /// Tile size for all-f32 operations.
-    pub f32_tile: u32,
-    /// Tile size for f16-input, f32-accumulator operations.
-    pub f16_tile: u32,
+    /// Default subgroup width, or zero if not exposed by the backend.
+    pub subgroup_size: u32,
```
kvark: This is not specific to coopmat, should be moved outside

blade-graphics/src/lib.rs
```diff
@@ -236,27 +236,32 @@ pub struct MemoryStats {
...
+    /// All-f32 operands, accumulator and result.
+    pub f32: Vec<[u32; 3]>,
```
kvark: Is the 3rd value derivable from the first 2?

kvark:
## Review notes

Verified the branch builds: `cargo check -p blade-graphics --lib --examples --offline` is clean. Also confirmed `maxThreadgroupMemoryLength` exists in objc2-metal 0.3.2 on `MTLDevice` (the feature is already enabled in `blade-graphics/Cargo.toml`), so the unbuilt Metal path should compile.

The core change is right. Replacing `find_tile`'s `m == n == k` filter with a shape list is exactly the fix, and the `matches!(s, 8 | 16)` bound correctly mirrors `CooperativeSize`'s per-dimension `Eight | Sixteen`. Two details worth calling out as good rather than as problems: adding `p.saturating_accumulation == vk::FALSE` is a correct tightening (a no-op for float configs today, but the right filter), and `sort_unstable` + `dedup` happens to preserve the old "prefer 8 over 16" ordering in the example, so there's no silent behavior change there.

Six things, roughly in severity order.

### 1. `subgroup_size` is unreachable on every device *without* cooperative matrix

`subgroup_size` is a device-wide property, but it's a field on `CooperativeMatrix`. All four early-out branches in `blade-graphics/src/vulkan/init.rs:529-566` return `CooperativeMatrix::default()`, so `subgroup_properties.subgroup_size` — which was queried successfully — is discarded on any device lacking the extension, the feature, or the memory model.

That's the wrong set. `blade-graphics/src/shader.rs:60-63` enables naga's `SUBGROUP | SUBGROUP_BARRIER` capabilities **unconditionally**, so subgroup shaders are legal everywhere. The devices that can't see the subgroup width are precisely the ones where cooperative matrix isn't in play but `subgroupAdd`/`subgroupBroadcast` still are. Given the stated goal — expose the subgroup width so callers can reject illegal kernel layouts before compilation — this defeats the purpose on most hardware.

Move it to `Capabilities` next to `max_compute_shared_memory_size`, and populate it before the coop-matrix branch.

### 2. `subgroupSize` is advisory, and the doc doesn't say so

Blade doesn't enable `VK_EXT_subgroup_size_control` and never requests a `requiredSubgroupSize` — the only mention in the tree is a comment at `shader.rs:55`. So `VkPhysicalDeviceSubgroupProperties::subgroupSize` is the device default, and the compiler may pick a different width per compute shader: Intel selects SIMD8/16/32, AMD RDNA can go wave32 or wave64.

A caller that sizes a cooperative-matrix kernel from this number can be wrong on exactly the Intel hardware this PR is motivated by. Either document it as advisory, or also expose `VkPhysicalDeviceSubgroupSizeControlProperties::{minSubgroupSize, maxSubgroupSize}` so callers see the range. "Default subgroup width" is technically accurate but reads as a guarantee in context.

### 3. GLES reports `0` shared memory while supporting compute

`blade-graphics/src/gles/mod.rs:478-479` sets `compute: self.capabilities.contains(Capabilities::BUFFER_STORAGE)` alongside `max_compute_shared_memory_size: 0`. GLES has real compute pipelines here (`gles/pipeline.rs:386`, `gles/command.rs:187`). A caller following the documented contract — zero means unknown, so be conservative — rejects *every* compute kernel on GLES. GLES 3.1 exposes `GL_MAX_COMPUTE_SHARED_MEMORY_SIZE` (guaranteed >= 16 KB), so this is a one-line `get_parameter_i32`.

### 4. The matmul example's fallback message is now misleading

On B570 the new code takes the fallback path (shapes exist, none square) and prints:

```
No square cooperative matrix supported on this device (Intel Arc B570).
Requires VK_KHR_cooperative_matrix (Vulkan) or Apple7+ (Metal).
```

The second line tells the user to obtain an extension the device already has and is reporting through. Before this PR that branch was only reachable when the extension was genuinely absent; now it's the normal outcome on the hardware that motivated the change. Suggest branching: if `cm.is_supported()` but no square shape, print the shapes found and say the example's generator only emits square tiles.

### 5. No CHANGELOG entry for a breaking API change

`docs/CHANGELOG.md` has an active `## (TBD)` section that already records this class of change (e.g. "`Capabilities::timing` tells whether the device can timestamp passes at all"). This PR renames `f32_tile`/`f16_tile` to shape vectors, adds `Capabilities::max_compute_shared_memory_size`, **and drops `Copy` from `CooperativeMatrix`** — that last one will break downstream code in ways the rename doesn't obviously telegraph. Worth two lines under TBD.

### 6. Naming

`cm.f32` and `cm.f16` as public field names are ambiguous — they read as element type, but the semantics are operand/accumulator combinations, which only the doc comments disambiguate. Since this is already a breaking change, it's the cheap moment to pick `f32_shapes` / `f16_f32_shapes` or similar.

Relatedly, `init.rs` now has `let f32 = find_shapes(f32t, f32t, f32t, f32t);`. It compiles (value and type namespaces are separate) but shadowing a primitive type name is worth avoiding.

---

Nothing here is a correctness bug in the Vulkan query itself — that part looks right and matches what B570 should report. #1 is the one I'd treat as blocking, since it makes the new subgroup field near-useless for its stated purpose.


---
_Generated by [Claude Code](https://claude.ai/code)_

blade-graphics/src/lib.rs
```diff
@@ -236,27 +236,46 @@ pub struct MemoryStats {
...
+    ///
+    /// On Vulkan this is `VkPhysicalDeviceSubgroupProperties::subgroupSize`.
+    /// Blade does not request a required subgroup size, so a compute shader
+    /// may run at another width. [`Self::min_subgroup_size`] and
+    /// [`Self::max_subgroup_size`] carry the `VK_EXT_subgroup_size_control`
+    /// range when the device provides it, including on Vulkan 1.3, and are
+    /// zero when that range is unknown.
+    pub subgroup_size: u32,
+    /// Smallest advertised subgroup size, or zero when the range is unknown.
+    pub min_subgroup_size: u32,
```
kvark: I don't think this is useful. Let's just have the `subgroup_size` alone for now.

kvark (commented):
CI is also unhappy

